PR #44, opened 2025-04-12
Small tweaks

corbt:
1. Sets the local python version to 3.10, the lowest we support. This should give us runtime/linter errors if we try to use features that aren't supported in this version.

2. Adds a new `Trajectory.for_logging` method. Includes the information I think we'll want to log to disk/logging providers like OpenPipe from a trajectory. Not used anywhere yet, but adding it since I'm using it in an outside project.

3. adds a `to_messages` convenience util. This is useful to convert a list of `MessagesAndChoices` back into just `messages`, so we can keep calling our agent in a loop. We should probably update our examples to use this where appropriate but want to see if it's a pattern others agree with first.


corbt (commented):
Also added `polars` as a dependency, since it looks like that's now imported in src/art/local/api.py

PR #54, opened 2025-04-16
Example benchmarking comparison models

corbt:


examples/benchmarking_comparison_models.py
```diff
@@ -0,0 +1,207 @@
...
+        gemini_flash,
+        gemini_flash_thinking,
+    ]
+
+    # We also can define the models we want to train.
+    qwen = art.TrainableModel(
+        name="qwen-2.5-14b-instruct",
+        project="my-project",
+        base_model="Qwen/Qwen2.5-14B-Instruct",
+        config=MyConfig(),
```
arcticfly, line 179: In practice I don't think I would provide an empty config. I'd probably just dynamically determine whether one was present in my rollout function, like this:

```
model_id = model.config.litellm_model_name if isinstance(model.config, CustomConfig) else f"openai/{model.name}"
```

src/art/utils/litellm_utils.py
```diff
@@ -0,0 +1,57 @@
+from openai.types.chat.chat_completion import Choice, ChoiceLogprobs
+from openai.types.chat.chat_completion_message import ChatCompletionMessage
+from openai.types.chat.chat_completion_token_logprob import (
+    ChatCompletionTokenLogprob,
+    TopLogprob,
+)
+from litellm.types.utils import Choices, ChoiceLogprobs as LitellmChoiceLogprobs
+
+
+def convert_litellm_choice_to_openai(litellm_choice: Choices) -> Choice:
```
arcticfly, line 10: @corbt Why is this necessary? Thus far I haven't run into issues using the default litellm choices.

corbt: One reason I did this is because it makes the typechecking work. But also I think there are some subtle things that might break if we pass in the wrong type, eg. looks like this line explicitly checks for an openai `Choice` type: https://github.com/OpenPipe/ART/blob/main/src/art/gather.py#L152

src/art/model.py
```diff
@@ -12,23 +12,71 @@
 )
 import httpx
 
+from openai import OpenAI
+
+client = OpenAI()
+
+client.base_url
```
bradhilton, line 19: I assume this isn't needed here?

examples/benchmarking_comparison_models.py
```diff
@@ -0,0 +1,209 @@
...
+# This function runs our rollout against our test scenarios, and logs the results.
+# We can use it to benchmark both trainable and prompted models.
+async def benchmark_model(model: art.Model):
+    trajectories = await art.gather_trajectories(
+        (rollout(model, scenario) for scenario in test_scenarios),
+        pbar_desc="benchmark",
+        max_exceptions=100,
+    )
+    valid_trajectories = [t for t in trajectories if isinstance(t, art.Trajectory)]
+    await model.log(valid_trajectories)
```
bradhilton, line 69: I want to change `log` to also accept exceptions so that this kind of filtering isn't necessary. Also, being able to log exceptions will be valuable.

examples/benchmarking_comparison_models.py
```diff
@@ -0,0 +1,209 @@
...
+import os
+
+PROJECT_NAME = "benchmarking-comparison-models"
+
+
+# This is a project-specific definition of a scenario you might use for training
+class MyScenarioSpec(BaseModel):
+    input_1: int
+    input_2: int
+    output: int
```
bradhilton, line 16: Nice. My preferred noun is 'Task,' but we still have no codified vocabulary for this in the library, right? Ultimately up to the user what they want to call this abstraction, but it's a good for the user to think about and can be domain specific (e.g. puzzle, game, question, assignment, etc.)

bradhilton (approved):
Beautiful example. LGTM.

examples/benchmarking_comparison_models.py
corbt: yes agreed!

examples/benchmarking_comparison_models.py
corbt: I don't feel too strongly about the word `scenario`.

My only concern with the word `task` is that it's pretty widely used to mean "a running job", eg. including in both [Skypilot](https://docs.skypilot.co/en/latest/reference/api.html#task) and [asyncio](https://docs.python.org/3/library/asyncio-task.html#creating-tasks). Since we use both those libraries heavily and the semantics of what we'd mean by `task` aren't the same as what the concepts refer to there (and other places), I worry about the potential for confusion.

I am at least weakly in favor of having one word we use across all our docs and standardize on a `rollout` function that only takes two args, `model` and (`task` or `scenario`), and returns a `Trajectory`. I think it's good to have a default pattern to help people new to the library build a mental model.

Admittedly though, that shape won't always work, eg. for self-play scenarios that @saum7800 is experimenting with we at a minimum need multiple trajectories returned from a rollout.

examples/benchmarking_comparison_models.py
```diff
@@ -0,0 +1,209 @@
...
+        inference_api_key=os.getenv("OPENAI_API_KEY"),
+    )
+
+    # We'll create a second GPT-4.1 variant that uses CoT reasoning
+    gpt_4_1_thinking = gpt_4_1.model_copy(deep=True)
+    gpt_4_1_thinking.name = "gpt-4.1-thinking"
+    assert isinstance(gpt_4_1_thinking.config, MyConfig)
+    gpt_4_1_thinking.config.use_thinking = True
+
+    gemini_flash = gpt_4_1.model_copy(deep=True)
```
arcticfly, line 158: I realize this is up to the user, but do you see a benefit to copying here over redeclaring a new gemini_flash model from scratch? Seems like the only preserved property is the project name.

examples/benchmarking_comparison_models.py
```diff
@@ -0,0 +1,209 @@
...
+
+    # We need to register all our models with the local API so they're
+    # available for training and logging.
+    api = art.LocalAPI()
+    for model in train_models + prompted_models:
+        await model.register(api)
+
+    # For prompted models, we can benchmark them right away.
+    for model in prompted_models:
+        await benchmark_model(model)
```
arcticfly, line 200: What do you think of encouraging the user to benchmark all models simultaneously? Same question for training multiple models simultaneously using SkypilotAPI. The difficulty of debugging an error might be higher (because it's less immediately clear what model was at fault, and logs might get confused across models) but the total run time should significantly decrease.

src/art/model.py
```diff
@@ -52,6 +100,38 @@ async def register(self, api: "LocalAPI") -> None:
         self._api = api
         await self._api.register(self)
 
+    def openai_client(
+        self,
+    ) -> AsyncOpenAI:
+        if self.inference_api_key is None or self.inference_base_url is None:
+            raise ValueError(
+                "OpenAI client not yet available. You must call `model.register()` first."
```
arcticfly, line 108: I don't believe this error message makes sense anymore, since comparison models' inference values aren't populated by the register command.

Maybe we should move this message to an overloaded function on `TrainableModel` that checks for an api key and base url and throws this error if they're not present, else calls super?

src/art/model.py
```diff
@@ -52,6 +100,38 @@ async def register(self, api: "LocalAPI") -> None:
...
+            )
+        openai_client = AsyncOpenAI(
```
arcticfly, line 110: Rather than recreating this client for each request, I'd rather use a singleton pattern (define the client once, check if its defined, reuse if present) to reduce CPU overhead and time spent on TLS handshakes.

(see my [reference implementation](https://github.com/OpenPipe/ART/pull/73/files#r2060619101))

corbt: Fair, updated to store. It's unclear to me whether there's any extra TLS handshakes from creating the client every time but possible there are.

src/art/model.py
corbt: 👍 updated to have different error messages for `Model` vs `TrainableModel`.

examples/benchmarking_comparison_models.py
corbt: Yeah makes sense, I've updated the script to work that way now. Slightly worried about readability for folks not super familiar with async/await but hopefully most people are.

PR #67, opened 2025-04-19
Helpers to load model files from S3 and push a model to S3.

corbt:
@bradhilton I haven't tested this yet, but wanted to open a PR in case you have feedback on the API shape before investing more time. Idea is to make it easy to save/load your full ART runs to S3, so you can preserve them if your machine crashes, you're using an ephermeral instance, or you want to set the machine to auto shut down when a run ends.

In the current implementation there's no way to pass s3 credentials, we're just assuming the user has set them via env vars.

It would be nice to support other storage sources like r2 / b2, but probably a lower priority.

cc @arcticfly @angkywilliam 

Usage would look something like:

```python
import art
import asyncio
from dotenv import load_dotenv
import json
import openai
import random
import re
from typing import TypedDict

load_dotenv()


class TemporalCluePuzzle(TypedDict):
    num_clues: int
    prompt: str
    solution: dict[str, str]


puzzles: list[TemporalCluePuzzle] = json.load(open("./data/temporal-clue/puzzles.json"))
val_puzzles = puzzles[:64]
test_puzzles = puzzles[64:128]
train_puzzles = puzzles[128:]
random.seed(42)
random.shuffle(train_puzzles)


model = art.TrainableModel(
    name="001",
    project="temporal-clue",
    base_model="Qwen/Qwen2.5-14B-Instruct",
    _internal_config={"init_args": {"gpu_memory_utilization": 0.775}},
)
api = art.LocalAPI()
await model.register(api)
await api._experimental_push_to_s3(
    model,
    s3_bucket=os.environ["BACKUP_BUCKET"],
)


async def rollout(
    client: openai.AsyncOpenAI, puzzle: TemporalCluePuzzle
) -> art.Trajectory:
    messages: art.Messages = [{"role": "user", "content": puzzle["prompt"]}]
    chat_completion = await client.chat.completions.create(
        messages=messages, model=model.name
    )
    choice = chat_completion.choices[0]
    content = choice.message.content
    assert isinstance(content, str)
    num_correct = 0
    for key, value in puzzle["solution"].items():
        if matches := re.findall(rf"{key}\. ([A-Za-z \.:-]+)", content):
            match = matches[-1]
            if match.strip().lower() == value.lower():
                num_correct += 1
    reward = acc = num_correct / len(puzzle["solution"])
    return art.Trajectory(
        messages_and_choices=[*messages, choice], reward=reward, metrics={"acc": acc}
    )


stride = 4
openai_client = model.openai_client()
for i in range(await model.get_step(), 1_000):
    val_groups, train_groups = await asyncio.gather(
        art.gather_trajectory_groups(
            (
                art.TrajectoryGroup(rollout(openai_client, puzzle) for _ in range(2))
                for puzzle in val_puzzles
            ),
            pbar_desc="val",
        ),
        art.gather_trajectory_groups(
            (
                art.TrajectoryGroup(rollout(openai_client, puzzle) for _ in range(50))
                for puzzle in train_puzzles[i * stride : (i + 1) * stride]
            ),
            pbar_desc="train",
        ),
    )
    await model.log(val_groups)
    await model.delete_checkpoints()
    await model.train(
        train_groups,
        config=art.TrainConfig(learning_rate=5e-5),
    )
    if i % 10 == 0:
            await api._experimental_pull_from_s3(
                model,
                s3_bucket=os.environ["BACKUP_BUCKET"],
            )

```

### DIFF
--- a/examples/hn_title_generator/train.py
+++ b/examples/hn_title_generator/train.py
@@ -19,7 +19,7 @@ BASE_MODEL = "Qwen/Qwen2.5-7B-Instruct"
 MAX_COMPLETION_LENGTH = 100
 MAX_PROMPT_LENGTH = 8192 - MAX_COMPLETION_LENGTH
 LEARNING_RATE = 1.2e-5
-ENTRIES_PER_STEP = 1
+GROUPS_PER_STEP = 1
 EVAL_STEPS = 50
 VAL_SET_SIZE = 100
 TRAINING_DATASET_SIZE = 5000
@@ -285,7 +285,7 @@ async def main():
 
     data_iterator = iterate_dataset(
         dataset=train_data_list,
-        batch_size=ENTRIES_PER_STEP,
+        groups_per_step=GROUPS_PER_STEP,
         num_epochs=NUM_EPOCHS,
         initial_step=start_step,
         use_tqdm=True,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ dependencies = [
     "tblib>=3.0.0",
     "litellm>=1.63.0",
     "polars>=1.26.0",
+    "awscli>=1.38.1",
 ]
 
 [build-system]
--- a/src/art/local/api.py
+++ b/src/art/local/api.py
@@ -32,6 +32,7 @@ from .tokenize import tokenize_trajectory_groups
 from .checkpoints import (
     delete_checkpoints,
     get_step,
+    get_last_checkpoint_dir,
 )
 from .s3_sync import s3_sync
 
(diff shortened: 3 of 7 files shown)


bradhilton:
Definitely a valuable feature. Dedicated pull/push methods to an external bucket may not make as much sense for a managed service, so we may prefer automatic, background synchronization. Optionally provide a bucket path in the `LocalAPI` initializer, automatically sync on model register and when changes are made to the model directory. Would also be nice to support arbitrary buckets. There may be libraries that can do this for us. I think only one parameter would be needed, `bucket_address` or `bucket_path` or `storage_path`, etc., which can include the scheme (e.g. "s3://"), bucket name, and a path within the bucket.

corbt:
> we may prefer automatic, background synchronization

Makes sense, that was the other API shape I was considering. I'll update this to look more like that!

PR #69, opened 2025-04-19
feat: Create a servable API

bradhilton:
This PR should allow users to train multiple models on H100s from the comfort of their own laptop. Provisioning GPUs, start the servers, and connecting properly will all be handled by SkyPilotAPI.

### Changes
* Separate local and remote API features
* Remove LocalAPI export from the top-level __init__.py file to avoid client-side dependency errors
* Add SkyPilotAPI to manage skypilot machines for ART run
* Run art server task on port 7999 instead of 2148 to make it closer to vLLM's default 8000
* Only require skypilot cluster name when initializing API

**NOTES:**

I haven't actually tested launching a skypilot server from a version of openpipe-art published on pypi, since we haven't yet published a version that includes the cli server command.


src/art/local/api.py
```diff
@@ -273,12 +273,14 @@ async def _train_model(
             )
             return
         disk_packed_tensors = packed_tensors_to_dir(
-            packed_tensors, f"{get_output_dir_from_model(model)}/tensors"
+            packed_tensors, f"{get_model_dir(self._path, model)}/tensors"
         )
         results: list[dict[str, float]] = []
-        pbar = tqdm.tqdm(total=disk_packed_tensors["num_sequences"], desc="train")
-        async for result in service.train(disk_packed_tensors, config, _config):
+        num_steps = disk_packed_tensors["num_sequences"]
```
arcticfly, line 279: Is this a different kind of step than the step that refers to a complete iteration through the training loop?

bradhilton: Yeah, these are gradient steps. We could rename this variable to `num_grad(ient)_steps` to clarify.

arcticfly: Will do!

src/art/__init__.py
```diff
@@ -10,18 +11,18 @@
...
 from .trajectories import Trajectory, TrajectoryGroup
 from .types import Messages, MessagesAndChoices, Tools, TrainConfig
-from .local import LocalAPI
 from .utils import retry
 
 __all__ = [
     "dev",
     "gather_trajectories",
     "gather_trajectory_groups",
-    "LocalAPI",
```
bradhilton, line 24: Removing the `LocalAPI` export is a breaking change.

bradhilton:
Some general feedback: changing the majority example notebooks from LocalAPI to SkyPilotAPI is pretty significant. LocalAPI only requires a GPU whereas SkyPilot set up is a bit more involved. I think for now we should maybe leave most of the example notebooks unchanged and provide some inline comments for any that we do change and/or additions to the README to explain SkyPilot usage.

src/art/__init__.py
arcticfly: Cool, can upgrade the major version!

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "openpipe-art"
-version = "0.1.25"
+version = "1.0.0"
```
bradhilton, line 3: I assume this is a typo. I wouldn't bump to 1.0.0, not for a while 😅

arcticfly:
> Some general feedback: changing the majority example notebooks from LocalAPI to SkyPilotAPI is pretty significant. LocalAPI only requires a GPU whereas SkyPilot set up is a bit more involved. I think for now we should maybe leave most of the example notebooks unchanged and provide some inline comments for any that we do change and/or additions to the README to explain SkyPilot usage.

I think 2048 still uses the LocalAPI class, which is the one which matters the most. Since SkyPilotAPI allows for faster client-side iteration and training multiple models simultaneously, I think we should definitely include examples of how to use it. But I can change back tic-tac-toe if you're worried about that one in particular.

pyproject.toml
arcticfly: No, I just incremented the major version because you pointed out this PR includes a breaking change. Did you have something else in mind? I think at this stage, since we're in alpha/beta, it's fine not to follow semver strictly.

bradhilton:
> I think 2048 still uses the LocalAPI class, which is the one which matters the most, since we link to it from our README. Since SkyPilotAPI allows for faster client-side iteration and training multiple models simultaneously, I think we should definitely include examples of how to use it. But I can change back yes-no-maybe if you're worried about that one in particular.

Yeah, I guess just change back `yes-no-maybe`.

pyproject.toml
bradhilton: Yeah, I just feel like anything >= 1.0.0 has expectations of being production ready. I'd just bump to `0.2.0`. I think most package managers pin major and minor versions and only allow the patch number to automatically upgrade, so I believe this should be okay.

PR #73, opened 2025-04-22
Launch cluster when initializing art api

arcticfly:


skypilot-config.yaml
```diff
@@ -383,6 +383,10 @@
 workdir: .
 resources:
   accelerators: ["H100:1", "A100-80GB:1"]
+  ports:
+    - 8080
```
arcticfly, line 387: Not sure if this port is necessary. It was exposed by default and I haven't tested without it.

src/art/cli.py
```diff
@@ -16,9 +17,20 @@
 
 
 @app.command()
-def run(host: str = "0.0.0.0", port: int = 2218) -> None:
+def run(host: str = "0.0.0.0", port: int = 7999) -> None:
```
angkywilliam, line 20: Where does the cli run being called on?

src/art/skypilot/api.py
```diff
@@ -0,0 +1,203 @@
...
+    _cluster_name: str
+
+    def __init__(
+        self,
+        *,
+        cluster_name: str = "art",
+    ) -> None:
+        self._cluster_name = cluster_name
+
+    async def initialize_cluster(
```
angkywilliam, line 27: Should we consider changing this to a static or class method?
- It feels a bit odd that `SkyPilotAPI.__init__` doesn't extend `API.__init__`, and instead `API.__init__` is called by `initialize_cluster`.
- Would it make sense to call `initialize_cluster` first, then pass the return output to `SkyPilotAPI.__init__`?

src/art/model.py
```diff
@@ -53,6 +55,33 @@ async def register(self, api: "API") -> None:
...
+        force_recreate: bool = False,
+    ) -> AsyncOpenAI:
+        if isinstance(self, TrainableModel) and (
+            self.base_url is None or self.api_key is None
+        ):
+            raise ValueError(
+                "OpenAI client not yet available. You must call `model.register()` first."
+            )
+
+        if self._openai_client is None or force_recreate:
```
arcticfly, line 69: Singleton allows us to skip TLS handshakes and reduces CPU overhead.

skypilot-config.yaml
arcticfly: Removed, was not necessary.

src/art/cli.py
arcticfly: The CLI process will be run on an H100 RunPod instance, and is the main art server that the client will make contact with when registering or training models.

src/art/skypilot/api.py
arcticfly: I don't have strong feelings here, but I do agree that it is a bit weird that `SkyPilotAPI`'s init function doesn't call `API`'s init function.

I'll move `initialize_cluster` and `_launch_cluster` to be outside of the `SkyPilotAPI` class.

arcticfly: Hmm, on second thought I don't love the idea of forcing the user to declare the name of the cluster twice (which you would need to do if you pass it in to `launch_cluster` and `__init__` separately.

```
    api = art.SkyPilotAPI(cluster_name="art6")
    await api.initialize_cluster(art_version=".", env_path=".env")
```

feels a lot cleaner to me than

```
    base_url = await initialize_cluster(cluster_name="art6", art_version=".", env_path=".env")
    api = art.SkyPilotAPI(cluster_name="art6", base_url=base_url)
```

src/art/skypilot/load_env_file.py
```diff
@@ -0,0 +1,22 @@
+from pathlib import Path
```
corbt, line 1: Why not just use python-dotenv for this and decrease our line count?

examples/tic_tac_toe/tic-tac-toe.py
```diff
@@ -0,0 +1,184 @@
...
+    return trajectory
+
+
+DESTROY_AFTER_RUN = False
+
+
+async def main():
+    # run from the root of the repo
+    api = art.SkyPilotAPI(cluster_name="art6")
+    await api.initialize_cluster(art_version=".", env_path=".env")
```
corbt, line 159: after looking at this I still feel it's too magical and I'd like us to add a required parameter named `gpu`, just like Modal has (if you want to run on GPU). I feel like as a user I'll have a much better mental model of what this line is doing if I explicitly have to put in that I'm provisioning a GPU here.

corbt: GPU could be optional if you pass in a `resources=` block instead.

src/art/__init__.py
```diff
@@ -11,19 +12,19 @@
...
-from .local import LocalAPI
 from .utils import retry
 
 __all__ = [
     "dev",
     "gather_trajectories",
     "gather_trajectory_groups",
     "API",
-    "LocalAPI",
+    "SkyPilotAPI",
```
corbt, line 27: I'm not sure we want to include this here/encourage people to import directly from the top-level `art` object. Skypilot is a heavy dependency that a lot of folks won't need; I'd rather not import it into the top-level object so in the future we can ship the core without it, and if someone does want to use our skypilot support they can run `uv add openpipe-art[skypilot]`.

skypilot-config.yaml
```diff
@@ -398,5 +401,9 @@ setup: |
   git clean -fd
   # Install astral-uv
   sudo snap install --classic astral-uv
+  # Install tmux
+  apt install tmux -y
+  # install awscli
+  uv add awscli
```
corbt, line 407: We should probably add this as a dependency of openpipe-art (I realize I should have done that).

src/art/__init__.py
```diff
@@ -34,3 +35,10 @@
     "Trajectory",
     "TrajectoryGroup",
 ]
+
+# Avoid importing LocalAPI on client machines
+if sys.platform.startswith("linux"):
```
corbt, line 40: This feels pretty ugly; maybe we should think harder about what should be imported from where.

src/art/skypilot/api.py
```diff
@@ -0,0 +1,203 @@
...
+            # check if art_version is valid semver
+            if art_version is not None:
+                try:
+                    semver.Version.parse(art_version)
+                    art_version_is_semver = True
+                except Exception:
+                    pass
+
+            if art_version_is_semver:
+                art_installation_command = f"uv pip install art=={art_version}"
```
corbt, line 124: nit: `uv add art=={art_version}` would probably work fine too, and give us an error if for some reason it didn't work with awscli or w/e.

src/art/skypilot/api.py
```diff
@@ -0,0 +1,203 @@
...
+            )
+            print("Art server task started")
+
+        art_endpoint = await to_thread_typed(
+            lambda: endpoints(cluster=self._cluster_name, port=7999)[7999]
+        )
+        base_url = f"http://{art_endpoint}"
+        print(f"Using base_url: {base_url}")
+
+        super().__init__(base_url=base_url)
```
corbt, line 94: Might be good to add a check here that we can actually connect to the remote port as expected. And maybe put it in a little loop that checks once a sec for 10 secs or something in case it takes a few seconds for the remote server to actually start listening.

corbt (commented):
Left a few comments, but looks pretty good and clean overall!

src/art/skypilot/api.py
arcticfly: Oh yeah, that seems better to me.

src/art/skypilot/api.py
arcticfly: It might work, but we won't know until we ship a version of art that supports the cli command.

skypilot-config.yaml
arcticfly: I think it is already added to ART, so we can remove this line from the script.

src/art/__init__.py
arcticfly: It is ugly. I think the problem might be solved by removing `LocalAPI` from the top `__init__.py` file. Just pushed that change.

src/art/skypilot/api.py
arcticfly: That's a good idea. Are you imagining a `/healthcheck` endpoint on the server to ping?

corbt: Yep, that's a good way to do it.

corbt (approved):

PR #74, opened 2025-04-23
Update yes-no-maybe example rollout function to accept model instead of client

arcticfly:
By accepting a model in our rollout function instead of a client parameter, we can support any kind of model that a user wishes to pass in, including custom attributes defined in its configuration. In the case of the example updated in this PR, the only attribute that we pass in is an optional `litellm_model_name` string, but other more complicated rollout functions might use attributes like `should_reason`.

The update to this rollout function does introduce a depedency on litellm and also requires the user to convert the litellm choice to OpenAI. It also misses out on the patching that we're doing in `patch_openai`. Is the patching still necessary @bradhilton? It looks like we'll be missing token counts if we stop relying on it, though it should be possible to calculate them in another way.

### Changes
* Accept model instead of client in yes-no-maybe rollout function
* Add gpt4o comparison model to yes-no-maybe example
* Use litellm to make completions
* Create and log validation trajectories using gpt4o


bradhilton:
Yes @arcticfly  I believe the only remaining usage of `patch_openai` is for token counts, both streaming total and average prompt/completion counts. The user can provide prompt and completion counts manually via metrics if they want to.

arcticfly:
Closing because we want to standardize on the default OpenAI SDK for now.

PR #81, opened 2025-04-29
Update skypilot, use async api

Kovbo:
Skypilot switched to a new client-server architecture where all client methods are async. Now we have a SkyPilot client and a SkyPilot server running on the same machine. The client just starts jobs, and the server executes them in the background. If we want to see logs, we need to capture them manually.


src/art/skypilot/api.py
```diff
@@ -57,37 +59,49 @@ async def initialize_cluster(
...
-            lambda: sky.status(cluster_names=[self._cluster_name])
+            lambda: sky.stream_and_get(sky.status(cluster_names=[self._cluster_name]))
         )
         if (
             len(cluster_status) == 0
             or cluster_status[0]["status"] != sky.ClusterStatus.UP
         ):
-            await self._launch_cluster(resources, art_version, env_path)
+            launch_job_id = await self._launch_cluster(resources, art_version, env_path)
+            await asyncio.to_thread(
```
arcticfly, line 69: I'd prefer to use the `to_thread_typed` helper for all our sync -> async conversions in this file. I realize that in the case of `tail_logs`, `to_thread` already properly returns the type, but I actually prefer the `to_thread_typed` syntax personally.

src/art/skypilot/api.py
```diff
@@ -34,7 +36,7 @@ async def initialize_cluster(
     ) -> None:
         self = cls.__new__(cls)
         self._cluster_name = cluster_name
-
+        self._job_id = None
```
arcticfly, line 39: Could we rename this to `_art_server_job_id`? We may have more tasks in the future, and I can imagine myself forgetting whether this id refers to the task of launching the cluster as a whole or the art server in particular.

src/art/skypilot/api.py
```diff
@@ -163,9 +186,13 @@ async def _launch_cluster(
         print(task)
 
         try:
-            await to_thread_typed(
-                lambda: sky.launch(task=task, cluster_name=self._cluster_name)
+            job_id, handler = await to_thread_typed(
```
arcticfly, line 189: Do we use handler anywhere?

src/art/skypilot/api.py
```diff
@@ -163,9 +186,13 @@ async def _launch_cluster(
...
+                lambda: sky.stream_and_get(
+                    sky.launch(task=task, cluster_name=self._cluster_name)
+                )
             )
+
+            return job_id
```
arcticfly, line 195: Instead of returning job_id here, would it make sense to start tailing logs within _launch_cluster? That would reduce the amount of code in the main initialize_cluster function, and seems like it should be part of the launching process.

uv.lock
arcticfly, line 1: Make sure to run `uv sync` on a linux machine. We've seen weird bugs from syncing on macbooks then running the backend on a linux machine.

arcticfly (commented):
Thanks for taking on this update!

This may be an issue with the new skypilot version (or an unrelated update), but when you launch a cluster through the SkyPilotAPI, do you see messages like these?

```
WARNING:google.auth.compute_engine._metadata:Compute Engine Metadata server unavailable on attempt 1 of 3. Reason: timed out
WARNING:google.auth.compute_engine._metadata:Compute Engine Metadata server unavailable on attempt 2 of 3. Reason: timed out
WARNING:google.auth.compute_engine._metadata:Compute Engine Metadata server unavailable on attempt 3 of 3. Reason: [Errno 64] Host is down
WARNING:google.auth._default:Authentication failed using Compute Engine authentication due to unavailable metadata server.
WARNING:google.auth.compute_engine._metadata:Compute Engine Metadata server unavailable on attempt 1 of 3. Reason: [Errno 64] Host is down
WARNING:google.auth.compute_engine._metadata:Compute Engine Metadata server unavailable on attempt 2 of 3. Reason: [Errno 64] Host is down
WARNING:google.auth.compute_engine._metadata:Compute Engine Metadata server unavailable on attempt 3 of 3. Reason: [Errno 64] Host is down
WARNING:google.auth._default:Authentication failed using Compute Engine authentication due to unavailable metadata server.
```

corbt:
@Kovbo when you get a chance do you want to check to make sure this still works and then we can get it in?

PR #87, opened 2025-04-30
Refactor s3, set default .art path

arcticfly:
This PR resolves a couple inconveniences I was experiencing:
* Multiple `.art` directories (depending on whether I was running a notebook or top-level python script)
* Difficulty getting logs onto my local machine for benchmarking
* Error when BACKUP_BUCKET has not yet been created

### Changes
* Move s3 logic out of local trainer and into `utils/s3.py` file.
* Default art directory to root of repo
* Create backup bucket if it does not already exist
* Look for BACKUP_BUCKET environment variable if no other name specified

Now in our benchmarking code, we can pull the model logs down like this:

```
from art.utils.benchmarking.load_trajectories import load_trajectories
from art.utils.s3 import pull_model_from_s3

await pull_model_from_s3(
    model_name="003",
    project=project_name,
)
await pull_model_from_s3(
    model_name="gpt-4o",
    project=project_name,
)
await pull_model_from_s3(
    model_name="gpt-4.1",
    project=project_name,
)
await pull_model_from_s3(
    model_name="gpt-4o-mini",
    project=project_name,
)

await load_trajectories.bust_cache()
df = await load_trajectories(
    project_name=project_name,
    models=["003", "gpt-4o", "gpt-4.1", "gpt-4o-mini"],
)
```


corbt:
Ok yeah this looks reasonable!

corbt (approved):

src/art/local/api.py
```diff
@@ -336,59 +342,48 @@ def _get_wandb_run(self, model: TrainableModel) -> Run | None:
...
+            verbose=verbose,
+            delete=delete,
+            art_path=self._path,
         )
-        await s3_sync(s3_path, local_dir, verbose=verbose, delete=delete)
 
         if isinstance(model, TrainableModel):
             service = await self._get_service(model)
-            lora_path = get_last_checkpoint_dir(local_dir)
+            lora_path = get_last_checkpoint_dir(local_model_dir)
```
arcticfly, line 367: @corbt Is the code in this if statement necessary? I would have expected `model.register` to take care of setting the service's LoRA.

corbt: This code is kinda hacky but is necessary unless we refactor in some other way to make things cleaner. The problem is as follows:

1. You need to register your model with the API before you can download it from s3
2. When you register a trainable model with the API, it checks what the checkpoint LoRA is and sets that as the one currently in use for inference
3. But you haven't downloaded the latest checkpoint from s3 yet! So the model will be set up to run inference with a new, untrained LoRA.

This code fixes it in a hacky way by re-setting the current LoRA after you've pulled from s3.

I think the right solution might be to split `register` and `prepare_for_inference` into separate methods (kinda like how @bradhilton had it originally, where `.openai_client()` prepared for inference). That way you can first `register`, then `pull_from_s3`, then `prepare_for_inference`  Or otherwise fold the reading-from-s3 into the `.register()` call so that it can be done, if necessary, before preparing the model for inference.

arcticfly: We could just remove the `_experimental_pull_from_s3` and `_experimental_push_to_s3` methods from the api altogether, and just use `pull_model_from_s3` and `push_model_to_s3` in our code, since at this point there isn't much difference anyway.

corbt: That would work with `LocalBackend`. Don't think it would work with `SkypilotBackend` though, since the push/pull has to happen on the remote backend?

arcticfly: As long as we register after pulling, should be fine.

PR #104, opened 2025-05-03
Easily track span duration within rollouts

corbt:
I ran into an issue on my current project where the llm-as-judge I was using was taking a super long time to return, bottlenecking the whole rollout process and leading to low GPU utilization.

This change makes it easy to track the duration of different parts of your rollout with a context manager. For example, to track how long an LLM-as-judge contained within a function takes, you could do the following:

```python
async with traj.track_duration("determine_if_answer_is_correct"):
    rubric.answer_correct = await determine_if_answer_is_correct(
        final_answer, scenario
    )
```

This will automatically add a new metric, `determine_if_answer_is_correct_duration`, which tracks the cumulative duration (in seconds) spent within the context manager.

I also add a new optional method to trajectories, `.finish()` When called, this will add a metric `duration` that tracks the total trajectory duration from start to finish.

Separately, this PR makes it explicit that metrics can be boolean or ints. This mostly already worked bc python will accept `int` in most places that expect `float`, and `bool` is a subclass of `int` in Python (!), but adding it to the types makes it more obvious.

Also fixes a minor type error in the art-e example. We should probably enable automatic typechecking as part of PRs at some point.

### DIFF
--- a/examples/art-e/README.md
+++ b/examples/art-e/README.md
@@ -44,7 +44,7 @@ To generate training data, the system:
 ```bash
 # Clone repository
 git clone https://github.com/OpenPipe/ART
-cd ART/examples/jarvis-mail
+cd ART/examples/art-e
 
 # Install package
 uv sync
--- a/examples/art-e/README.md
+++ b/examples/art-e/README.md
@@ -44,7 +44,7 @@ To generate training data, the system:
 ```bash
 # Clone repository
 git clone https://github.com/OpenPipe/ART
-cd ART/examples/art-e
+cd ART/examples/jarvis-mail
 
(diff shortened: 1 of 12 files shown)


examples/art-e/README.md
```diff
@@ -44,7 +44,7 @@ To generate training data, the system:
 ```bash
 # Clone repository
 git clone https://github.com/OpenPipe/ART
-cd ART/examples/art-e
+cd ART/examples/jarvis-mail
```
arcticfly, line 47: Is this name change intentional?

arcticfly (approved):

Issue #112, opened 2025-05-07
Training script hangs after job is done


arcticfly:
When running LocalBackend on a RunPod instance, I'm seeing a very long (possibly eternal?) hang after a training script theoretically finishes. Sample code below, taken from the [tic-tac-toe-local.py](examples/tic_tac_toe/tic-tac-toe-local.py) example.

```
DESTROY_AFTER_RUN = False


async def main():
    print(0)
    # run from the root of the repo
    backend = LocalBackend()

    print(1)

    model = art.TrainableModel(
        name="agent-001",
        project="tic-tac-toe-agent",
        base_model="Qwen/Qwen2.5-3B-Instruct",
    )
    print(2)
    await backend._experimental_pull_from_s3(model)
    print(3)
    await model.register(backend)
    print(4)

    step = await model.get_step()
    print(f"Step: {step}")

    for i in range(await model.get_step(), 101):
        train_groups = await art.gather_trajectory_groups(
            (
                art.TrajectoryGroup(
                    rollout(model, i, is_validation=False) for _ in range(200)
                )
                for _ in range(1)
            ),
            pbar_desc="gather",
        )
        await model.delete_checkpoints()
        await model.train(train_groups, config=art.TrainConfig(learning_rate=1e-4))
        await backend._experimental_push_to_s3(model)

    print(5)

    # res = await backend._experimental_deploy(model=model, verbose=True)
    # print(res)

    print(6)

    if DESTROY_AFTER_RUN:
        await backend.down()


if __name__ == "__main__":
    asyncio.run(main())
```

End of stdout:

```
Unsloth: Will smartly offload gradients to save VRAM!
train: 100%|██████████████████████████████████████████| 16/16 [00:18<00:00,  1.15s/it, loss=0.358, grad_norm=0.219, policy_loss=0.358]
5
6
```

stdout includes the number 6, indicating that everything has finished, but the script hangs anyway.

arcticfly:
This also occurs after a training script finishes as a result of an exception.

corbt:
@JonesAndrew are you taking a look at this?

bradhilton:
Has been addressed by @JonesAndrew's PR #126 🥇

PR #126, opened 2025-05-19
Fix hanging

JonesAndrew:
Make sure the training script shuts down cleanly. You either need to call .close or use the `with` syntax.

https://github.com/OpenPipe/ART/issues/112

### DIFF
--- a/src/art/local/backend.py
+++ b/src/art/local/backend.py
@@ -67,8 +67,13 @@ class LocalBackend(Backend):
         return self
 
     async def __aexit__(self, *excinfo):
-        for name, service in self._services.items():
-            service.close()
+        self.close()
+    
+    def close(self):
+        for _, service in self._services.items():
+            close_method = getattr(service, "close", None)
+            if callable(close_method):
+                close_method()
 
     async def register(
         self,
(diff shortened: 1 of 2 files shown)


src/mp_actors/move.py
```diff
@@ -199,7 +245,12 @@ async def _handle_request(
...
             result = result_or_callable
         response = Response(request.id, result, None)
+    except StopAsyncIteration:
+        generators.pop(request.id, None)
+        return
     except Exception as e:
-        pickling_support.install(e)
+        import traceback
+        traceback.print_exc()
+        pickling_support.install()
```
bradhilton, line 254: Is printing the traceback here necessary, or was this just for debugging?

src/art/local/backend.py
```diff
@@ -63,6 +63,18 @@ def __init__(self, *, in_process: bool = False, path: str | None = None) -> None
...
+
+    def __exit__(self, *excinfo):
+        self.close()
+    
+    def close(self):
+        for _, service in self._services.items():
+            close_method = getattr(service, "close", None)
+            if callable(close_method):
+                close_method()
+
```
bradhilton, line 77: It doesn't appear that this context manager or the close method are used anywhere so we can probably remove them, right?

bradhilton (changes requested):

src/art/local/backend.py
bradhilton: Oh never mind, this is the main feature of this PR. Got it 👍

src/mp_actors/move.py
JonesAndrew: Sorry yeah that was just for debugging

bradhilton (approved):

PR #128, opened 2025-05-20
Add Wiki-Game as an example

Kovbo:
I could not build a good-looking benchmark graph because my reward graph is too spiky. On wandb, I had to set smoothing to the maximum to see a pattern. Can I do it with our charts?


corbt:
Is it too spiky even with a large test set? If so sounds like training might not actually be converging?

Kovbo:
Each rollout's difficulty level is very different, as pages are selected randomly. Sometimes, a path between two pages is short, and a model can easily find it. In other cases, a short path does not exist. 
It should not affect training, but we can only see a trend by averaging results from multiple games.

WanDB can visualize the trend by smoothing the line graph, but the ART benchmark does not do it, or I don't know how to set the smoothing.
image

Perhaps, I could do a bar chart by averaging multiple games.

corbt:
Can you create a larger validation set and run it every N iterations? That should help us see whether the model is actually getting stronger on OOD tasks.

PR #135, opened 2025-05-21
feat: Add force_restart option to SkypilotBackend

corbt:
- Introduced a `force_restart` parameter to the SkyPilotBackend class to force cancellation of existing art server tasks if needed.
- Also added logic to sync local ART code to SkypilotBackend every time a new task is run, not just on cluster initialization.

Does not fix the hanging training jobs, but might make it easier to debug them.

### DIFF
--- a/src/art/skypilot/backend.py
+++ b/src/art/skypilot/backend.py
@@ -32,6 +32,7 @@ class SkyPilotBackend(Backend):
         resources: sky.Resources | None = None,
         art_version: str | None = None,
         env_path: str | None = None,
+        force_restart: bool = False,
     ) -> "SkyPilotBackend":
         self = cls.__new__(cls)
         self._cluster_name = cluster_name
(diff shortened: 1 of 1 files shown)


src/art/skypilot/backend.py
```diff
@@ -88,6 +98,11 @@ async def initialize_cluster(
             art_server_task.set_resources(cast(sky.Resources, resources))
             art_server_task.update_envs(self._envs)
 
+            # If a local path was provided for art_version, ensure it is mounted so the latest
+            # code is synced to the remote cluster every time we (re)launch the art_server task.
+            if art_version is not None and os.path.exists(art_version):
+                art_server_task.set_file_mounts({"~/sky_workdir": art_version})
```
arcticfly, line 104: Hmm, not sure we want to do this by default. It's conceivable that I might be editing code on the cluster itself, and I don't want that getting overwritten.

Maybe make this a parameter, and default to False?

arcticfly (approved):

src/art/skypilot/backend.py
corbt: As discussed we'll experiment with keeping this on by default for now and can turn it off if it ends up being annoying for @arcticfly's workflow.

PR #140, opened 2025-05-23
Fix training stability issues with new vLLM version

saum7800:
https://github.com/vllm-project/vllm/pull/12622 -- since this commit in vLLM, if you don't pass in a generation-config, it uses whatever it finds in generation_config.json from the model if it exists. if you want to use vllm defaults, you have to explicitly pass in generation_config="vllm". which is what used to happen by default before this commit.

For RL training, we need
repetition_penalty = 1.0
top_p = 1.0
top_k = 0
temperature = 1

Changes to the above changes the logprobs returned from vLLM which we use to calculate losses and gradient updates, which leads to unstable training.

we're setting the default generation config to "vllm" to have the above sampling params, instead of the generation_config.json from the model.


bradhilton (approved):
🚀

PR #150, opened 2025-06-03
feat: Add vLLM prometheus observability support

bradhilton:


bradhilton:
I like the approach in #177 better

PR #163, opened 2025-06-25
feat: Multi-device training

bradhilton:


Danau5tin:
Hi @bradhilton , thanks for notifying me!
Yes I would love to try it out, I really want to train Qwen3-32B, and normally would use deepspeed zero 3 for that and split the inference and training over multiple GPUs.

Am I right in saying (quick look), that ART doesn't split GPUs between training and inference. Instead, it uses all GPUs sequentially?
So if you have 8 GPUs, all 8 are used for inference, then all 8 for training, cycling back and forth as needed?

Do you think an A100 (80GB) is fine for Qwen3-32B?

Danau5tin:
Hey @bradhilton, so I migrated my project over to ART, tried out the code, and all looked great!
Then I realised the model wasn't training.

Here are some logs I have:
```
 Step 3: Generating trajectories: 100%|██████████████████████████████| 4/4 [01:51<00:00, 27.99s/it, reward=0.104, turns=3, test_reward=0, judge_reward=0.297, prompt_tokens=2.85e+3, total_completion_tokens=9386.0]
  
2025-07-03 18:29:15,391 - src.train_with_art - INFO - Collected 4 which have rewards: [0.147, 0.0875, 0.1295, 0.0525]. Stnd dev: 0.04257, prompt_tokens=2.85e+3, total_completion_tokens=9386.0]
  
Skipping tuning as there is no suitable data. This can happen when all the trajectories in the same group have the same reward and thus no advantage to train on.
```
Created with this code:
```python
        all_rewards = [
            trajectory.reward for group in train_groups for trajectory in group.trajectories
        ]
        logger.info(
            f"Collected {len(all_rewards)} which have rewards: {all_rewards}. Stnd dev: {torch.std(torch.tensor(all_rewards)):.4f}"
        )
```

Not sure where that warning at the bottom comes from, but it comes up no matter what the rewards in the group are, and I can see from other logs that the trajectories are actually happening as expected.

So to see if it was because of the multi-gpu setup, I switched to single GPU (removed this internal config below - just commented it out):
```
    internal_config = art.dev.InternalModelConfig(
        engine_args=art.dev.EngineArgs(
            tensor_parallel_size=torch.cuda.device_count(),
            gpu_memory_utilization=0.9,
        ),
        torchtune_args=art.dev.TorchtuneArgs(
            model=torchtune_model,
            model_type=model_type,
        ),
    )
```

When I switched to the standard way of training on a single GPU, it just seemed to hang on the first rollout. So not sure what is happening there either because the rollouts worked fine with the multi-gpu setup. 

It has been a long day, so only tried the single GPU setup - with Unsloth once before I wrote this. 

I am using Qwen 0.6B for reference (Just to try it out), on 2x A100 GPUs.

PR #167, opened 2025-06-26
Refactor dev/tau-bench for true async

saum7800:


dev/tau-bench/tau_bench/envs/user.py
```diff
@@ -41,10 +41,9 @@ def __init__(self, model: str, provider: str) -> None:
         self.model = model
         self.provider = provider
         self.total_cost = 0.0
-        self.reset()
```
saum7800, line 44: does this removal of self.reset() have any negative effects?

PR #170, opened 2025-06-28
feat: Add support for multiple histories

bradhilton:


src/art/local/tokenize.py
```diff
@@ -57,13 +57,27 @@ def tokenize_trajectory_groups(
...
                 continue
-            if result := tokenize_trajectory(
-                tokenizer,
-                trajectory,
-                advantage,
-                allow_training_without_logprobs,
-            ):
-                results.append(result)
+            trajectory_results: list[TokenizedResult] = []
+            for history in [
```
corbt, line 61: Probably doesn't matter but we could make `Trajectory` inherit from `History` to avoid having to create a new one here.

corbt (commented):
I do think it would be good to have an explicit way to mark an assistant message as trainable or not. I guess we get that right now by passing `Choice` vs `Message` objects but if we end up standardizing on using `Message` objects everywhere (or someone wants to pass `allow_training_without_logprobs` for any reason) we'll have an issue. That can come later though I suppose.

Also, just checking that these are being masked out somewhere so the different histories in the same trajectory don't attend to each other?

bradhilton:
@corbt no masking, here's a caveat I shared in our discussion:
> For the qwen 3 reasoning use case you'll want to add an additional message and choices list for each history. You'll also want to convert previous choice objects to plain assistant messages so that you don't multiplicatively train on earlier messages with the wrong contents to boot
I also mused "maybe we could add support for auto-splitting reasoning trajectories once we figure out what works," but I'm not ready to tackle that atm

bradhilton:
@corbt yes, each history will yield it's own `TokenizedResult`, which gets a unique token group id in the packing step

corbt (approved):
Sweet, excited to have this in. @arcticfly we'll definitely have to document how this works.

PR #173, opened 2025-06-29
Match default SkyPilotBackend version to client

### LINKED ISSUE #172: Client and server versions out of sync
When initializing SkyPilotBackend, the latest version of ART is used for the server if no particular version is specified and no path to the source code is provided. This can cause issues when the latest version of the server is no longer compatible with the installed version of the client.

We should instead default the server to whatever version of `openpipe-art` is installed on the client.

arcticfly:
Resolves #172 

### Changes
* During SkyPilotBackend initialization, default ART server to whatever version of `openpipe-art` is specified in client
* Throw error if `openpipe-art` not installed and no art_version provided

### DIFF
--- a/src/art/skypilot/backend.py
+++ b/src/art/skypilot/backend.py
@@ -4,6 +4,7 @@ import sky
 import os
 import semver
 from dotenv import dotenv_values
+from importlib.metadata import version, PackageNotFoundError
 
 from .utils import (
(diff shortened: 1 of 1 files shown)


arcticfly:
Merging to get into `0.3.12`, nothing appears to be broken after testing.

PR #177, opened 2025-06-30
Adds vllm metrics to wandb

corbt:
This PR starts tracking several vllm metrics to ART dashboards in wandb:
 - num_requests_running
 - num_requests_waiting
 - e2e_request_latency_seconds_p50
 - e2e_request_latency_seconds_p99

The metrics are collected from the vllm /metrics endpoint, which is exposed by the vllm
server. The metrics are collected every 15 seconds and logged to wandb.

This is helpful to see where the bottleneck is in your system. By watching the `num_requests_running` graph, for example, you can see whether your system is being bottlenecked by straggles.

Unfortunately to get the resolution I was looking for (1 sample per 15 seconds, similar to wandb's default system metrics) I had to stop tracking existing metrics against the `step` directly. Instead I'm now tracking them against a new `training_step` metric. Sadly, this means that runs from before and after this PR will no longer appear on the same axes. I'm not sure whether there's a better way to handle this.

Screenshot 2025-06-30 at 12 24 02 AM

Screenshot 2025-06-30 at 12 24 25 AM

Screenshot 2025-06-30 at 12 24 42 AM


bradhilton (commented):
I wonder if there's a way to provide custom wandb "system" metrics on a timestamp basis, so as to not mess with the default "step" @corbt

bradhilton:
Also, looks like there are big changes to a couple uv.lock files with no visible changes in dependencies, so I'm wondering if those changes can be safely discarded.

bradhilton (approved):

PR #178, opened 2025-06-30
Track entropy at training time

corbt:
It may be useful to see how entropy changes over time, since once entropy collapses the model is unlikely to learn much more.

Not sure whether this is the best way to do it; notably, when doing it this way we aren't able to track entropy on the val set, only at training.

This is a subset of my changes from the `increase-entropy` branch, where I tried to explicitly reward increased entropy. However that branch, while successfully maintaining higher entropy, didn't seem to help training, so I only copied over the reporting changes here.


bradhilton (approved):
Looks great!

PR #184, opened 2025-07-01
Wikihop with group judge

arcticfly:
### Changes
* Add wikihop/wikihop.ipynb notebook in examples
* Add wikihop/judge_group.py with generalized group judgement function


corbt:
Does it work? If the model isn't training properly we should put it in dev/ for now, not examples/

PR #185, opened 2025-07-01
feat: Add vLLM V1 support w/Unsloth model service

bradhilton:


bradhilton:
There are a few current limitations with [Unsloth Zoo](https://github.com/unslothai/unsloth-zoo) that disallow V1 support. Generally, Unsloth Zoo does not support V1's [collective RPC pattern yet](https://github.com/unslothai/unsloth-zoo/blob/816109fb7b3eed49442d667041d2d35baaaf2f5d/unsloth_zoo/vllm_utils.py#L612). The collective RPC call to get the weight IPC handles failed with `CUDA error: invalid argument`. Also, the collective RPC calls do not check if the results are coroutines and so fail when called from `AsyncLLM` instances.

corbt:
I'm not seeing any chatter on the Unsloth side about working towards this. How hard would it be to do it ourselves?

bradhilton:
Hard to say, could take a while.

bradhilton:
Probably will end up closing this if decoupling vLLM & Unsloth works out

PR #193, opened 2025-07-02
Log tools and additional histories

corbt:
Small change to add tools and additional histories to exported logs.

### DIFF
--- a/src/art/utils/trajectory_logging.py
+++ b/src/art/utils/trajectory_logging.py
@@ -3,16 +3,24 @@ from typing import Any
 
 from art import TrajectoryGroup, Trajectory
 from art.types import Choice, Message, MessageOrChoice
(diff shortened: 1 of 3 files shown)


src/art/utils/trajectory_logging.py
```diff
@@ -25,21 +33,32 @@ def trajectory_group_to_dict(trajectory_group: TrajectoryGroup) -> dict[str, Any
...
+        message_or_choice_to_dict(message_or_choice)
+        for message_or_choice in trajectory.messages_and_choices
+    ]
 
     return {
         "reward": trajectory.reward,
         "metrics": trajectory.metrics,
         "metadata": trajectory.metadata,
         "messages_and_choices": messages_and_choices,
+        "tools": trajectory.tools,
```
corbt, line 48: These two lines are the only real change.

arcticfly: Separate question: should this be combined with the `for_logging` function in trajectories.py?

PR #194, opened 2025-07-02
Move training/inference dependencies to [backend] group

corbt:
This change makes it so you can more easily install the ART client without installing the heavy-weight backend dependencies.

### DIFF
--- a/dev/tau-bench/run_training.py
+++ b/dev/tau-bench/run_training.py
@@ -109,7 +109,7 @@ def launch_model(model_key: str):
 
             # Install project in editable mode
             uv remove openpipe-art
-            uv add --editable ~/ART
+            uv add --editable ~/ART --extra backend
 
             # Sync dependencies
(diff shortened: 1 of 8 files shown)


corbt:
@bradhilton I've now tested this works as expected and updated the docs. Ready for review.

@arcticfly I didn't update any of our notebooks since they pin ART to 0.3.11 which doesn't have this change. But when we update them we'll need to change `pip install openpipe-art==0.3.11` to `pip install openpipe-art[backend]==0.3.14`.

arcticfly:
> when we update them we'll need to change pip install openpipe-art==0.3.11 to pip install openpipe-art[backend]==0.3.14.

Got it. Right now we're still failing to run openpipe-art>=0.3.12 on T4s, so we won't be able to do that update until we determine the cause of the issue.

bradhilton (approved):
LGTM

PR #221, opened 2025-07-10
ruler docs

corbt:
adding to the docs


docs/fundamentals/ruler.mdx
```diff
@@ -0,0 +1,279 @@
...
+import art
+from art.rewards import ruler_score_group
+
+# Create a TrajectoryGroup from your trajectories
+group = art.TrajectoryGroup([...])  # List of art.Trajectory objects
+
+# Use RULER to score them
+judged_group = await ruler_score_group(
+    group,
+    {"model": "openai/o3"},
```
arcticfly, line 49: Why nest model in a dictionary, instead of making it a top-level parameter?

corbt: lets us customize other litellm params, which is useful for eg. using a local model with a specific api endpoint.

corbt: that said, `model` is frequently going to be the only thing you need to customize. I'll make that its own explicit argument and have a separate one for additional litellm params.

arcticfly (approved):
Awesome!

PR #235, opened 2025-07-14
docs: update SkyPilotBackend installation instructions

corbt:
## Summary
- Updated documentation to include instructions for installing the `openpipe-art[skypilot]` optional dependency
- This change reflects the new package structure where SkyPilotBackend is behind an optional dependency group

## Changes
- Added installation instructions (`pip install openpipe-art[skypilot]`) to:
  - Installation + Setup guide
  - ART Backend documentation
  - Summarizer tutorial

## Test plan
- [x] Ran `./scripts/run_checks.sh` - all checks pass
- [x] Reviewed all documentation files mentioning SkyPilotBackend
- [x] Verified example projects don't need updates (they handle imports conditionally or use local development mode)

🤖 Generated with [Claude Code](https://claude.ai/code)


dev/test/main.py
```diff
@@ -0,0 +1,3 @@
+from art.skypilot import SkyPilotBackend
+
+backend = SkyPilotBackend()
```
arcticfly, line 3: What is this testing? Just imports, or actual cluster initialization? I think we should adjust the file name to reflect what it tests.

pyproject.toml
```diff
@@ -25,12 +25,13 @@ backend = [
     "setproctitle>=1.3.6",
     "weave>=0.51.51",
     "tblib>=3.0.0",
-    "semver>=3.0.4",
     "setuptools>=78.1.0",
     "wandb>=0.19.8",
     "polars>=1.26.0",
 ]
 
+skypilot = ["semver>=3.0.4", "skypilot==0.8.0"]
```
arcticfly, line 33: @Kovbo In our dev-dependencies block, we have this form of skypilot installed:

```
"skypilot[cudo,do,fluidstack,gcp,lambda,paperspace,runpod]==0.8.0"
```

Is it important that we include all of these extras?

arcticfly: @corbt We should also get rid of the skypilot dependency in dev-dependencies block.

dev/test/pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[project]
+name = "test"
```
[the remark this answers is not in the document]
corbt: Tbh I was just using this as a scratchpad and figured I'd commit it since I've had to create identical projects a few times in the past to check whether imports work. It would be good to have actual tests here but that can be a separate PR.

PR #243, opened 2025-07-15
Gemma 3 fix

Lucas-Fernandes-Martins:
As noted in the main READ.me, Gemma 3 models are not yet supported by ART, due to Gemma not accepting the enable_prefix_caching parameter.

To solve this issue, I've introduced the following changes on get_model_config.py:

```
use_gemma_config = config.get("use_gemma_config", False)

if use_gemma_config:
        init_args = InitArgs(
            model_name=base_model,
            max_seq_length=32768,
            load_in_4bit=True,  # False for LoRA 16bit
            fast_inference=True,  # Enable vLLM fast inference
            # vLLM args
            disable_log_stats=False,
            gpu_memory_utilization=(
                0.79 if enable_sleep_mode else 0.55
            ),  # Reduce if out of memory
            max_lora_rank=8,
            use_async=True,
        )
 else:
        init_args = InitArgs(
            model_name=base_model,
            max_seq_length=32768,
            load_in_4bit=True,  # False for LoRA 16bit
            fast_inference=True,  # Enable vLLM fast inference
            # vLLM args
            disable_log_stats=False,
            enable_prefix_caching=True,
            gpu_memory_utilization=(
                0.79 if enable_sleep_mode else 0.55
            ),  # Reduce if out of memory
            max_lora_rank=8,
            use_async=True,
        )
```

I believe this would solve the problem, as users can then specify the parameter use_gemma_config and avoid enable_prefix_caching to be added to the arg list.

Let me know if this is not correct or require adaptations. 

Thank you very much :)


bradhilton:
@Lucas-Fernandes-Martins have you been able to test this? Does it work?

bradhilton:
Thank you @Lucas-Fernandes-Martins for your investigation. I am afraid that adding Gemma 3 support will likely be tricky.

bradhilton:
Thank you @Lucas-Fernandes-Martins for investigating!

PR #253, opened 2025-07-17
feat: Add checkpoint forking functionality

corbt:
## Summary

This PR adds the ability to fork checkpoints from existing models, which is useful when training goes off the rails and you want to restart from a previous checkpoint with different parameters.

## Changes

- Added `_experimental_fork_checkpoint` method to Backend and LocalBackend classes
- Changed S3 pulling API from `latest_only` boolean to `only_step` parameter that accepts either an int or "latest"
- Added debugging output to help diagnose S3 sync issues

## Example Usage

```python
# In project_types.py, configure fork settings
class ProjectPolicyConfig(BaseModel):
    fork_from_model: str | None = None
    fork_from_project: str | None = None  
    fork_not_after_step: int | None = None

# In all_experiments.py, create a forked model
models["229"] = models["224"].model_copy(deep=True)
models["229"].name = "email-agent-229"
models["229"].config.fork_from_model = "email-agent-224"
models["229"].config.fork_not_after_step = 1381

# In train.py, fork will happen automatically if configured
if model.config.fork_from_model:
    await backend._experimental_fork_checkpoint(
        model,
        from_model=model.config.fork_from_model,
        from_s3_bucket=os.environ["BACKUP_BUCKET"],
        not_after_step=model.config.fork_not_after_step,
        verbose=True,
    )
```

### only_step parameter on 
- `only_step=None` - Pull all checkpoints (default)
- `only_step="latest"` - Pull only the latest checkpoint
- `only_step=1234` - Pull only checkpoint 1234

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,3 +13,19 @@ This project uses the `uv` package manager.
 ## Releases
 
 - If asked to help with a release, refer to the checklist in CONTRIBUTING.md
+
+## Documentation
+
+- All documentation is in the `docs` directory.
+- If you add a new page, be sure to add it to the sidebar in `docs/docs.json`.
+- If you move a page, be sure to update the sidebar in `docs/docs.json` and check for any broken links.
+
+### Adding images
+
+- Add images to the `docs/images` directory
+- If the image is a png, first convert it to webp using `magick <input.png> <output.webp>`. Do not include the original png in the repo.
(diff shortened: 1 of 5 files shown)


corbt:
The good news is that forking seems to work. (The bad news is that my forked run failed in almost the same place, but you can't win 'em all).

Screenshot 2025-07-16 at 5 48 22 PM

PR #292, opened 2025-08-02
Add configurable timeout for vLLM server initialization

### LINKED ISSUE #281: Timeout issue
In `art.vllm.server.py` file, the function `test_client()` often reaches timeout limit of the allowed 10 seconds during model initialization. Unfortunately this happens very often, and by manually increasing this limit to higher values (I set it to 5 minutes) I practically resolve all the timeout exceptions occuring when loading the model via calling `backend.register()`

It would be good if that timeout could be controlled through a parameter (or ENV variable), or even if you would increase this value manually to some other higher constant.

Some basic sysinfo:
```
Python 3.12.11
openpipe-art 0.4.4
GPU: 3x NVIDIA A40
CPUs: 96x Intel(R) Xeon(R) Gold 6336Y CPU @ 2.40GHz
```

geekalaa:
Fixes #281 by making the OpenAI server startup 
  timeout configurable via the ART_SERVER_TIMEOUT environment variable.

  ## Summary
  - Add ART_SERVER_TIMEOUT environment variable support
  - Improve error message to inform users about the timeout option  
  - Maintain backward compatibility with 10-second default

  ## Test plan
  - [x] Verified environment variable parsing works correctly
  - [x] Confirmed backward compatibility with default 10s timeout
  - [x] Tested error message includes helpful guidance

  ## Future work
  Add environment variables documentation section to docs.

### DIFF
--- a/src/art/vllm/server.py
+++ b/src/art/vllm/server.py
@@ -78,7 +78,9 @@ async def openai_server_task(
             return_when="FIRST_COMPLETED",
         )
         if not done:
-            raise TimeoutError(f"Unable to reach OpenAI-compatible server within {timeout} seconds. You can increase this timeout by setting the ART_SERVER_TIMEOUT environment variable.")
+            raise TimeoutError(
+                f"Unable to reach OpenAI-compatible server within {timeout} seconds. You can increase this timeout by setting the ART_SERVER_TIMEOUT environment variable."
+            )
         for task in done:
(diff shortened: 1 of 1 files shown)


arcticfly:
@geekalaa looks like the code quality checks failed. Could you run `./scripts/run_checks.sh --fix` to fix formatting?

Issue #271, opened 2025-07-26
Could I add a translated Readme.md file in Chinese(Simplified)?


8086CJ:
I am a native Chinese speaker. I will try my best.

PR #297, opened 2025-08-05
Add a translated version of README.md in Chinese

8086CJ:
README_zh_CN.md is the translated version of README.md in Chinese(Simplified).

This close issue #271.


corbt:
Thanks for the contribution! Do you know if there's a good automated flow we could adopt to keep the two README versions in sync? I worry about content drift.

PR #298, opened 2025-08-05
Add MCP•RL

arcticfly:
### Changes
* Add notebook
* Link notebook in README
* Replace RULER intro in README with MCP•RL section


src/art/skypilot/backend.py
```diff
@@ -55,7 +55,7 @@ async def initialize_cluster(
         if resources is None:
             resources = sky.Resources(
                 cloud=sky.clouds.RunPod(),
-                accelerators={"H100": 1},
```
corbt, line 58: Probably a bad idea since a lot of providers won't have that listed as an explicit type

PR #314, opened 2025-08-07
Fix GBK codec error when reading pyproject.toml

linpan:
Added encoding="utf-8" to file read/write operations in bump_version.py to handle pyproject.toml with UTF-8 encoding, resolving 'gbk' codec can't decode byte 0xbf error at position 1779.


bradhilton:
@linpan what's the error you're seeing? also looks like you may need to run `scripts/run_checks.sh --fix`

bradhilton:
I don't see  "#中国" “中国” in `pyproject.toml`

bradhilton:
@linpan can you run `scripts/run_checks.sh --fix`? once the format issue is fixed then i can merge

PR #320, opened 2025-08-07
fix: set openai library upperbound to 1.99.1

batjandra:
With the release of GPT-5, OpenAI seemed to have introduced breaking changes to their oai library. For instance, doing 

`from art.model import Model` 

will lead to 

`cannot import name 'Function' from 'openai.types.chat.chat_completion_message_tool_call'`. 

Setting the upper bound to 1.99.1 fixes the above issue.


bradhilton:
Thanks @batjandra!

PR #322, opened 2025-08-08
Cleanup jupyter notebook logs

Kovbo:
Our examples have unnecessary logs from previous runs. Some of them contain crashes and errors. We can remove them.

Separately, I wanted to use the latest version of ART in our examples, but realized it doesn’t work on T4 GPUs.
The vLLM v1 engine doesn’t plan to support them, and vLLM v0.10.0 drops support for the v0 engine.
We should consider other options for hosting our examples, or possibly come up with a workaround.
@bradhilton, is there a way to use vLLM 0.9.0 for examples without messing with the codebase and still being on the latest art version?


bradhilton:
oh, interesting. are you sure it drops support for v0? i thought I've still been seeing v0 in the logs.

perhaps we support 0.9.2 through 0.10.0 and pin it to 0.9.2 in the colab notebooks

Kovbo:
They say they have already [started removing v0 in v0.10](https://github.com/vllm-project/vllm/issues/18571):
> v0.9: Last supported version of the V0 engine, with frozen features and continued testing.
> v0.10: Version marking the start of V0 code removal.
> v0.11: First version without any V0 components.

It would be good if we could keep supporting v0.9 just for Colab.

PR #353, opened 2025-08-21
Add `just-the-facts` example

arcticfly:
This is a pretty basic example of a Qwen 2.5 14B model learning to give unbiased summaries of news articles, similar to SummaryRL. Nothing groundbreaking, but it does show how to use several LLM-based checks in conjunction with one another.

### Changes
* Add `just-the-facts` example
* Allow percent bar chart to include percents with decimals


bradhilton (approved):
NICE!

PR #354, opened 2025-08-22
Update notebooks

Kovbo:
We can run the latest version of ART in Google Colab!

Also, Google Colab recently updated its version of PyTorch, so we need an extra installation step similar to how [Unsloth does it](https://colab.research.google.com/github/unslothai/notebooks/blob/main/nb/Qwen3_(4B)-GRPO.ipynb#scrollTo=VMOgDrPItgoO).


bradhilton (approved):
LGTM

PR #372, opened 2025-08-27
Pin gql verion to fix 

Kovbo:
`gql` v4 breaks Weave launch during model registration:
```
"/root/sky_workdir/.venv/lib/python3.10/site-packages/weave/wandb_interface/wandb_api.py"
SyncClientSession.execute() takes 2 positional arguments but 3 were given
```
Let's pin gql as we do in notebook examples.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "openpipe-art"
-version = "0.4.10"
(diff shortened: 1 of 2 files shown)


corbt:
looks like due to https://github.com/wandb/weave/issues/5288

PR #388, opened 2025-09-02
`LocalBackend._monitor_openai_server` improvement

giladfrid009:
- Do not add additional load the server while performing health-check by sending inference requests; rather check if the actual model is loaded on the server via `openai_client.models.retrieve`
- added configurable timeout for the health check via env variable `"ART_SERVER_MONITOR_TIMEOUT"`


giladfrid009:
@bradhilton requesting review :)

PR #426, opened 2025-10-05
Add client-side error capturing

Kovbo:
Fixed an issue where failed serverless training jobs would never report a failure to the client.
Errors are now thrown and logged to the console.


bradhilton (approved):
Good addition!

PR #446, opened 2025-10-28
ENH: Improve assertion error message for content type

freddiev4:
Enhance assertion for content type in message dictionary.

Came across this when testing this out with a tool-call and the assertion / why it was failing was not obvious


arcticfly (approved):
Thanks @freddiev4!

freddiev4:
> Thanks @freddiev4!

np! Looks like you’ll need to merge as I don’t have permission

PR #509, opened 2026-01-15
fix(ruler): Add explicit JSON schema format for Ollama models in RULER

### LINKED ISSUE #476: RULER + Litellm with Ollama: `ollama_chat` recommended, but `/api/chat` cannot generate required JSON schema
**Description**

When configuring Ollama via Litellm, the Litellm docs recommend using:

> `ollama_chat` for better responses

This uses Ollama’s `/api/chat` endpoint.

However, in ART’s RULER integration, the judge model needs to:

* Produce **JSON-structured outputs** (e.g. match a Pydantic schema / tool schema) so RULER can parse correctness, reasoning, etc.

The problem:

* With **Ollama’s `/api/chat`** (via `ollama_chat`), the responses are not reliably:

  * Valid JSON matching the expected schema, or
  * Tool-call-compatible in the way RULER expects.
* As a result:

  * The judge’s responses often fail schema validation.
  * RULER cannot extract the required fields, and the scoring fails.

Because of that, I’m forced to fall back to the older **`/api/generate`** endpoint (`ollama`), even though:

* Litellm explicitly recommends `ollama_chat` over `ollama`.
* `/api/chat` is the more modern endpoint.

**What I expect**

* Either:

  * Official support / documentation for using **Ollama’s `/api/chat`** (`ollama_chat`) as a RULER judge with JSON-schema / tool-call style responses; or
  * Clear guidance that for RULER’s JSON-schema needs, we must currently use `/api/generate` and `ollama`.

**What actually happens**

* Using `ollama_chat`:

  * The `/api/chat` endpoint does not produce the JSON / tool-call schema RULER expects.
  * Judge calls fail schema validation.
* Using `ollama`:

  * RULER works better, but we lose out on the newer `/api/chat` behavior Litellm recommends.

**Request**

* Please:

  * Document the supported Ollama configuration for RULER (which engine, which endpoint, any special settings).
  * If possible, add direct support for `ollama_chat` + `/api/chat` that:

    * Ensures JSON-mode / tool-call style output works with RULER’s structured response expectations.
  * Or provide example configs + templates that make Ollama’s `/api/chat` usable with RULER.

siddikisahil47:
## Summary
Fixes #476

Adds support for [ollama_chat/](cci:1://file:///e:/Sahil/open-source/ART/tests/unit/test_ruler.py:18:4-20:65) and [ollama/](cci:1://file:///e:/Sahil/open-source/ART/src/art/rewards/ruler.py:52:0-54:74) models in RULER by converting the Pydantic [response_format](cci:1://file:///e:/Sahil/open-source/ART/src/art/rewards/ruler.py:57:0-69:5) to the explicit JSON schema format that LiteLLM requires for Ollama's `/api/chat` endpoint.

## Changes
- Added helper to detect Ollama models
- Added helper to convert Pydantic schema to explicit JSON schema dict
- Updated [ruler()](cci:1://file:///e:/Sahil/open-source/ART/src/art/rewards/ruler.py:72:0-244:28) to use correct format based on model type
- Added unit tests

## Testing
- ✅ All unit tests pass
- ✅ Verified manually with `ollama_chat/qwen2.5:7b`

### DIFF
--- a/src/art/rewards/ruler.py
+++ b/src/art/rewards/ruler.py
@@ -50,6 +50,26 @@ DEFAULT_RUBRIC = dedent(
 as RULER extracts task understanding from the system prompts in the trajectories."""
 
 
+def _is_ollama_model(model: str) -> bool:
+    """Check if the model is an Ollama model."""
+    return model.startswith("ollama/") or model.startswith("ollama_chat/")
+
+
+def _get_response_format_for_ollama(pydantic_model: type[BaseModel]) -> dict:
+    """Convert Pydantic model to explicit JSON schema format for Ollama.
+
+    LiteLLM requires an explicit JSON schema format for Ollama models to
+    produce structured output via the /api/chat endpoint.
+    """
+    return {
+        "type": "json_schema",
+        "json_schema": {
+            "name": pydantic_model.__name__,
+            "schema": pydantic_model.model_json_schema(),
+        },
+    }
+
+
 async def ruler(
     message_lists: list[list[ChatCompletionMessageParam]],
     judge_model: str = "openai/o3",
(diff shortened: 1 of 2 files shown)


siddikisahil47:
Closing this PR

Following up on the feedback from @corbt – I've confirmed that this fix belongs at the LiteLLM level.

I verified that with the latest LiteLLM (v1.80.17), RULER now works perfectly with Ollama models on the main branch without needing this workaround at the caller level.

To ensure this remains fixed, I've contributed regression tests directly to the LiteLLM repository: [BerriAI/litellm#19269](https://github.com/BerriAI/litellm/pull/19269).

Thank you for the guidance! This also resolves #476 in this repository

PR #525, opened 2026-01-20
SFT preprocessing

angkywilliam:


bradhilton (changes requested):
can we move the unsloth zoo import into the `tokenize_sft_batches` function? importing unsloth zoo at the file level may make this file unusable on macOS (for Tinker usage, for example).

PR #526, opened 2026-01-20
Fix OpenAI temperature handling

corbt:
Allow temperature=0.0 to pass through to Tinker sampling params by only defaulting to 1.0 when unset.


bradhilton (commented):
looks like there's an issue with the quality checks

PR #530, opened 2026-01-22
WIP: SFT (local backend)

Kovbo:
screencapture-localhost-3002-fundamentals-sft-training-2026-02-16-15_27_52


src/art/utils/model_config.py
```diff
@@ -0,0 +1,71 @@
+"""Model-specific configuration for chat templates and training defaults."""
+
+
+def detect_chat_template_parts(
```
angkywilliam: We need to add support to handle `OpenPipe/Qwen3-14B-Instruct`?

src/art/utils/sft.py
```diff
@@ -0,0 +1,399 @@
...
+    if verbose:
+        print(f"Loaded {len(trajectories)} trajectories from {file_path}")
+
+    if len(trajectories) == 0:
+        if verbose:
+            print("No trajectories to train on")
+        return
+
+    # Prepare dataset: shuffle per epoch, concatenate, and calculate learning rates
+    all_trajectories, learning_rates = prepare_sft_dataset(
```
angkywilliam: I might be missing something here. Won't this approach cause memory to blow up for large file?
E.g 1GB file train for 10 epoch will take at least 10GB itself outside of the system overhead itself.

src/art/unsloth/service.py
```diff
@@ -434,6 +463,180 @@ async def train(
...
+        # Reload training model to GPU (after vLLM is asleep)
+        self._state.reload_to_gpu()
+
+        # Get model and optimizer
+        peft_model = self._state.peft_model
+        self._reset_optimizer_if_mode_changed("sft")
+        optimizer = self._state.trainer.optimizer
+
+        # Set SFT-specific hyperparameters
+        sft_weight_decay = 0.01
```
angkywilliam: Should we set this to 0 to be consistent with existing behavior in OpenPipe Platform?

src/art/unsloth/service.py
```diff
@@ -434,6 +463,180 @@ async def train(
...
+                batch_loss += loss.item()
+
+            # Gradient clipping
+            grad_norm = torch.nn.utils.clip_grad_norm_(
+                peft_model.parameters(), max_grad_norm
+            ).item()
+
+            # Optimizer step at the end of each batch
+            optimizer.step()
+            optimizer.zero_grad()
```
angkywilliam: Remove optimizer.zero_grad()
We have clear the grad before the training loop

src/art/unsloth/service.py
```diff
@@ -434,6 +463,180 @@ async def train(
...
+
+        # Get model and optimizer
+        peft_model = self._state.peft_model
+        self._reset_optimizer_if_mode_changed("sft")
+        optimizer = self._state.trainer.optimizer
+
+        # Set SFT-specific hyperparameters
+        sft_weight_decay = 0.01
+        for param_group in optimizer.param_groups:
+            param_group["weight_decay"] = sft_weight_decay
```
angkywilliam: Should we set adam_beta1 and adam_beta2 as well?

src/art/utils/sft.py
```diff
@@ -0,0 +1,292 @@
...
+    if row_count == 0:
+        if verbose:
+            print("No trajectories to train on")
+        return
+
+    # Calculate total trajectories and batches
+    total_trajectories = row_count * epochs
+    skip_trajectories = initial_step * batch_size
+
+    if skip_trajectories >= total_trajectories:
```
angkywilliam: Remove skip_trajectories?
I don't think we support continuous training with our current design

src/art/unsloth/service.py
Kovbo: Yeah, good point! Changed.

src/art/unsloth/service.py
Kovbo: We change weight_decay because it is different for SFT and RL. But adam beta is the same, so there is no need to override it.

src/art/utils/model_config.py
Kovbo: The `detect_chat_template_parts` function should already handle these models:
```
    if "<|im_start|>" in template:
        return "<|im_start|>user\n", "<|im_start|>assistant\n"
```

I would prefer the custom MODEL_CONFIGS list to contain only unusual chat templates that cannot be resolved by `detect_chat_template_parts`.

src/art/utils/sft.py
Kovbo: I just added the final_step suuport, and this way it should work with continuous training.
A user can break down training on a large dataset into several calls, and benchmark a checkpoint after each step. 
It is just two arguments of the `train_sft_from_file` util function that help to correctly split the dataset and calculate LR rates.

docs/fundamentals/sft-training.mdx
```diff
@@ -0,0 +1,238 @@
...
+        }
+      }
+    }
+  ]
+}
+```
+
+Each line must follow these rules:
+
+- **`messages`** (required) — a non-empty list of chat messages. Each message has a `role` (`system`, `user`, `assistant`, `developer`, `tool`, or `function`) and `content`. The last message **must** be from the `assistant` role.
```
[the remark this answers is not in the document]
Kovbo: We do not explicitly validate roles, it passes messages through to the tokenizer's apply_chat_template, so what's supported depends on the base model's chat template. So yeah, let's not document `function`, and `developer`.

docs/fundamentals/sft-training.mdx
```diff
@@ -0,0 +1,238 @@
...
+    # backend = ServerlessBackend()  # or use serverless for managed GPUs
+    model = art.TrainableModel(
+        name="warmup-then-rl",
+        project="my-project",
+        base_model="Qwen/Qwen3-30B-A3B-Instruct-2507",
+    )
+    await model.register(backend)
+
+    # Phase 1: SFT warmup from a dataset
+    await train_sft_from_file(
```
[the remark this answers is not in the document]
arcticfly: Could you document that here, specifying that the model will train for 49 steps in the following loop?

src/art/utils/sft.py
```diff
@@ -0,0 +1,508 @@
...
+        method=schedule_type,
+        warmup_steps=warmup_steps,
+    )
+
+    # chunk_size is in batches; compute trajectory count per chunk
+    items_per_chunk = batch_size * chunk_size
+    chunks_per_epoch = math.ceil(dataset_size / items_per_chunk)
+
+    # Convert initial_step (batch-based) to initial_chunk for skipping
+    initial_chunk = initial_step // chunk_size
```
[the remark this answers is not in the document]
Kovbo: Both `initial_step` and `chunk_size` are already in batch units:                                          
`initial_step` is: "Global batch step to resume from"                                        
`chunk_size` is: "Number of batches to process per train_sft call"                             
                                                                                                                                         
  So initial_step // chunk_size = batches / batches-per-chunk = chunk index.

PR #554, opened 2026-02-12
Require model.get_inference_name() and fail loudly on model.name

corbt:
## Summary
- migrate rollout/inference call sites from `model.name` to `model.get_inference_name()` across examples/dev scripts/docs
- remove unsuffixed inference alias behavior so trainable model names must include `@step`
- make TinkerNative return clear 400 errors for missing/invalid `@step` suffixes
- align LocalBackend server health checks with step-qualified inference names
- update related multi-checkpoint unit coverage

## Why
Recent behavior allowed unsuffixed `model.name` to silently route to stale step-0 adapters in some backends. This change makes incorrect usage fail loudly and standardizes explicit `model@step` inference routing.

## Validation
- `uv run --extra backend pytest tests/integration/test_multi_checkpoint_training.py::test_local_backend -q`
- `uv run --extra tinker pytest tests/integration/test_tinker_native_backend.py -q`
- `uv run pytest tests/unit/test_multi_checkpoint_inference.py -q -k "served_model_name_uses_base_model_when_lora_enabled or TestStepParsing"`
- yes-no-maybe smoke runs for both LocalBackend and TinkerNativeBackend showed eval improvement over run


arcticfly (approved):
Thanks. Do you want me to update the notebooks as well, or are you already working on that?

PR #625, opened 2026-03-20
Replace kl_penalty_reference_step with kl_penalty_step_lag

angkywilliam:
## Summary

  - Rename `kl_penalty_reference_step` to `kl_penalty_step_lag` in PipelineTrainer
  - `None` (default): uses step 0 as KL reference (anchor to initial model)
  - `>= 1`: uses `max(0, current_step - lag)` as reference (rolling anchor)
  - Add validation that `kl_penalty_step_lag` must be >= 1 if specified

  ## Test plan

  - [x] Unit tests pass (`test_pipeline_trainer_local_backend.py`)
  - [x] New tests for lag computation added
  - [x] Integration tests (requires 2 GPUs)

### DIFF
--- a/dev/run_yes_no_maybe_kl_advantage_tinker.py
+++ b/dev/run_yes_no_maybe_kl_advantage_tinker.py
@@ -1,0 +1,104 @@
+"""Launch yes-no-maybe-kl-advantage-tinker training on SkyPilot (Kubernetes).
+
+Usage:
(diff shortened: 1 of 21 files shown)


arcticfly:
I'd prefer to default to no kl penalty if no argument is provided, and use checkpoint 0 if step lag of 0 is provided.

PR #635, opened 2026-03-31
feat: Add dedicated lora mode to Megatron backend

vivekkalyan:
## Summary

This adds dedicated `lora` mode to the Megatron backend so inference and training can run on separate GPUs in parallel.

In dedicated mode, ART now keeps a dedicated vLLM server on the inference GPU and updates LoRA adapters in place after training steps. That makes the Megatron flow match the dedicated-serving model we already use elsewhere instead of treating training as a blocking operation.

## What this enables

- Use Megatron with separate trainer and inference GPUs in dedicated mode
- Keep rollout serving on the inference GPU while Megatron trains on the trainer GPU
- Advance served model steps by reloading the next adapter into the running vLLM server

## Validation

- Unit coverage:
  - `tests/unit/test_megatron_dedicated.py`
- 2-GPU smoke on a fresh H200 machine:
  - trainer on GPU `0`
  - inference on GPU `1`
  - base model `Qwen/Qwen3-30B-A3B-Instruct-2507`
  - dedicated `lora` mode completed two real train steps and advanced the served model from `@0` to `@2`

### DIFF
--- a/src/art/megatron/backend.py
+++ b/src/art/megatron/backend.py
@@ -1,3 +1,5 @@
+import os
+
 from mp_actors import move_to_child_process
 
 from ..local.backend import LocalBackend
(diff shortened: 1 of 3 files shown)


vivekkalyan:
Rebased this PR onto the latest `main` changes, including the Megatron training fix from #646. I reran the dedicated Megatron validation on the rebased branch and confirmed the dedicated LoRA path still works and performs a real training step.

PR #642, opened 2026-04-04
Expose LocalBackend from top-level package

NenoL2001:
## Summary
- expose `art.LocalBackend` via lazy top-level export
- keep backend-only dependencies optional during `import art`
- add a regression test for the top-level export

## Notes
- no runtime behavior change beyond restoring the documented top-level API

### DIFF
--- a/src/art/__init__.py
+++ b/src/art/__init__.py
@@ -17,8 +17,12 @@ command above and they will appear in the project directory.
 """
 
 import os
+from typing import Any, TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


Kovbo:
Should we instead standardize on using `from art.local import LocalBackend` and remove `art.LocalBackend` from public API? I think it matches our examples.
[PR](https://github.com/OpenPipe/ART/pull/717)
cc @bradhilton

PR #653, opened 2026-04-16
fix S3 presign URI for checkpoint archive deployment

yu2001-s:
## Summary
- pass the full `s3://...` URI to `aws s3 presign` when archiving checkpoint steps
- add focused unit coverage for the archive upload + presign command path
- verify presign failures still surface the AWS CLI stderr

## Root cause
`archive_and_presign_step_url()` stripped the `s3://` prefix before invoking `aws s3 presign`. The AWS CLI expects an S3 URI there, so Together deployment flows could fail after a successful archive upload because the presign command was handed an invalid path.

## Validation
- `uv run pytest tests/unit/test_s3_utils.py -q`
- `uv run ruff check src/art/utils/s3.py tests/unit/test_s3_utils.py`
- `uv run --with safetensors ty check src/art/utils/s3.py tests/unit/test_s3_utils.py`

## Notes
- `uv run pytest --nbval --current-env --tb=short tests/unit -q` is not runnable in this default local environment because several existing unit-test modules require optional heavy deps (`torch`, `transformers`, `datasets`) that are not installed.

### DIFF
--- a/src/art/utils/s3.py
+++ b/src/art/utils/s3.py
@@ -322,11 +322,8 @@ async def archive_and_presign_step_url(
         await ensure_bucket_exists(s3_bucket)
         await s3_sync(archive_path, s3_step_path, verbose=verbose, delete=delete)
 
-    # Remove the s3:// prefix to get the key
-    s3_key = s3_step_path.removeprefix("s3://")
-
     # Generate presigned URL with 1 hour expiration
-    cmd = ["aws", "s3", "presign", s3_key, "--expires-in", "3600"]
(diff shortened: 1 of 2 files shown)


Kovbo:
Thanks for the PR, but  I could not reproduce this as a real failure.
Both AWS CLI v1.45.23 and AWS CLI v2.34.62 accept the stripped form:

```
aws s3 presign demo-bucket/path/file.zip
aws s3 presign s3://demo-bucket/path/file.zip
```

Both commands exited successfully and generated presigned URLs for the same bucket/key.

Issue #656, opened 2026-04-22
PipelineTrainer + LocalBackend dedicated mode hangs at training step 2


arcticfly:
## Summary

When using `PipelineTrainer` with `LocalBackend` in dedicated mode (`trainer_gpu_ids`/`inference_gpu_ids` set), training consistently hangs after completing step 1. Both GPUs drop to 0% utilization and the process never advances to step 2.

## How to reproduce

Branch: `test-forked-v5`

```bash
uv run dev/yes-no-maybe-fork-pipeline.py
```

Or on a 2-GPU cloud node via SkyPilot:

```bash
sky launch dev/yes-no-maybe-fork-pipeline.sky.yaml --env WANDB_API_KEY=<key>
```

**Requirements:** 2 GPUs (H100/H200/A100-80GB), `meta-llama/Llama-3.1-8B-Instruct` access.

## Observed behavior

```
Training base model from step 0 → 10...
  Eval at step 0: 70.8%
[23:01:36] INFO service.py:617: [DEDICATED] _train_dedicated: inference weights updated for step 1
<hangs indefinitely — both GPUs at 0% utilization>
```

Step 1 completes successfully (rollouts → train → checkpoint save → adapter reload), but step 2 never starts. Tested across 6 separate job runs; all exhibit the same hang.

## What does NOT hang

- **`yes-no-maybe-fork.py`** (same yes/no/maybe task and model, same checkpoint forking, but using a plain manual training loop instead of `PipelineTrainer`) — **works correctly** through 10 steps + fork + 2 more steps. This confirms the underlying model, rewards, and `LocalBackend` dedicated mode are all functional; the hang is specific to `PipelineTrainer`.
- The integration test `tests/integration/test_pipeline_localbackend_dedicated.py` (`max_steps=2`, `min_batch_size=1`, `eval_fn=None`, Qwen3-0.6B) — **passes**.

## Suspected location

Inside `run_unsloth_rl_training` in `src/art/unsloth/train.py`, one of two places:

1. **`await ctx.results_queue.join()`** at the start of each step — if `_unfinished_tasks > 0` because GRPOTrainer called `log()` more times than expected for step 1, this would block forever.
2. **`asyncio.wait([results_queue.get(), ctx.train_task], ...)`** — if neither task completes in the context of the nested event loop (via `nest_asyncio`).

## Diagnostic logging added

[...]

PR #657, opened 2026-04-22
dev: Add checkpoint forking end-to-end tests

arcticfly:
## Summary

- Adds `dev/yes-no-maybe-fork.py` and its `sky.yaml`: end-to-end test using a plain training loop with `LocalBackend` dedicated mode. Trains Llama 3.1 8B for 10 steps, forks the checkpoint into a fresh model, verifies the fork starts at step 10, and trains 2 more steps. **Confirmed working** (job SUCCEEDED).
- Adds `dev/yes-no-maybe-fork-pipeline.py` and its `sky.yaml`: same test but using `PipelineTrainer`. Currently hangs at training step 2 — tracked in issue #656.
- Adds diagnostic logging to `run_unsloth_rl_training` to help identify the `PipelineTrainer` hang (see #656).

## Test plan

- [ ] `uv run dev/yes-no-maybe-fork.py` completes successfully (already verified)
- [ ] `uv run dev/yes-no-maybe-fork-pipeline.py` completes successfully (blocked on #656)
- [x] `uv run prek run --all-files`

### DIFF
--- a/src/art/pipeline_trainer/trainer.py
+++ b/src/art/pipeline_trainer/trainer.py
@@ -767,7 +767,8 @@ class PipelineTrainer(Generic[ScenarioT, ConfigT]):
     async def _log_zero_variance_groups(self, step: int) -> None:
         if not self._discard_queue:
             return
-        discarded = list(self._discard_queue)
+        # Cap logged groups to avoid huge parquet files when many discards accumulate.
+        discarded = list(self._discard_queue[:50])
         await self.model.log(discarded, split="discarded", step=step)
         self._discard_queue.clear()
 
--- a/src/art/pipeline_trainer/trainer.py
+++ b/src/art/pipeline_trainer/trainer.py
@@ -4,8 +4,8 @@ import asyncio
 import os
 import signal
 import time
-import warnings
 from typing import Any, AsyncIterator, Generic, Iterable, TypeVar, cast
+import warnings
 
 T = TypeVar("T")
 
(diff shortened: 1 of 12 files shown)


src/art/unsloth/train.py
```diff
@@ -820,20 +829,56 @@ async def run_unsloth_rl_training(
...
+            )
 
-            done, _ = await asyncio.wait(
+            done, pending = await asyncio.wait(
                 [
                     asyncio.create_task(ctx.results_queue.get()),
                     ctx.train_task,
                 ],
                 return_when=asyncio.FIRST_COMPLETED,
+                timeout=300.0,
```
vivekkalyan: is this timeout absolutely required to not have issues with training? what was the previous behavior that this is trying to address?

my concern is if this might affect inadvertently affect larger training steps.

src/art/pipeline_trainer/trainer.py
```diff
@@ -134,7 +136,14 @@ def __init__(
         self.max_steps = max_steps
         self._status_log_interval_seconds = log_interval_seconds
         self.eval_every_n_steps = eval_every_n_steps
-        self.eval_step_0 = eval_step_0
+        if eval_step_0 is not None:
```
vivekkalyan: my preference here is to not have backward compatibility and let the user's script fail early since I don't feel we are bound to this API contract

vivekkalyan (approved):

src/art/unsloth/train.py
arcticfly: Addressed in d57c4dcf: removed the fixed 300s timeout so `asyncio.wait` returns only when a result arrives or the train task finishes/raises, matching the previous no-timeout behavior while keeping cleanup of the abandoned queue get task.

src/art/pipeline_trainer/trainer.py
arcticfly: works for me

PR #660, opened 2026-04-24
fix: resolve ruler function/module name collision in art.rewards (#339)

### LINKED ISSUE #339: Bug: art.rewards.ruler function and module collision
```bash
Art version: 0.4.4
Python: 3.12.11
```

the function `art.rewards.ruler.ruler()` and the module itself `art.rewards.ruler` have the same name. 
Since in the `art.rewards.__init__` you import the function `art.rewards.ruler.ruler()` as `"ruler"`, a collision occurs between the module name and the function.

To demonstrate:

(1)

```python
import art.rewards.ruler as art_ruler
print(art_ruler.DEFAULT_RUBRIC)
```

(2)

```python
from art.rewards import ruler as art_ruler
print(art_ruler.DEFAULT_RUBRIC)
```

(3)

```python
import art.rewards.ruler
print(art.rewards.ruler.DEFAULT_RUBRIC)
```

All Resulting in the error:

```bash
AttributeError: 'function' object has no attribute 'DEFAULT_RUBRIC'
```

Basically, I dont even think it's possible to somehow import `DEFAULT_RUBRIC`, or anything which is defined inside the 
`art.rewards.ruler` module while not being imported in `art.rewards.__init__`.

fixing it will be a breaking change either way I believe, so I would suggest renaming the function `art.rewards.ruler.ruler()` to something else, as most likely those who use RULER probably use it from the `ruler_score_group()` function.

robrui:
### Problem

The import `from .ruler import ruler` in `art.rewards.__init__` shadowed the `art.rewards.ruler` module name with the `ruler()` function, making the module-level constants (`DEFAULT_RUBRIC`, `Response`, `TrajectoryScore`, etc.) inaccessible via `import art.rewards.ruler` or `from art.rewards import ruler as art_ruler`.

Attempting to access `DEFAULT_RUBRIC` through the module raised:
```
AttributeError: function object has no attribute DEFAULT_RUBRIC
```

### Fix

Changed the import pattern from:
```python
from .ruler import ruler, ruler_score_group
```
to:
```python
from . import ruler as _ruler
ruler = _ruler.ruler
ruler_score_group = _ruler.ruler_score_group
```

This keeps both the module accessible (via `_ruler` internals) and the existing public API (`art.rewards.ruler()`, `art.rewards.ruler_score_group()`) intact — no breaking changes.

Fixes #339

### DIFF
--- a/src/art/rewards/__init__.py
+++ b/src/art/rewards/__init__.py
@@ -1,3 +1,5 @@
-from .ruler import ruler, ruler_score_group
+from . import ruler as _ruler  # noqa: F811 — keep module accessible for DEFAULT_RUBRIC, Response, etc.
+ruler = _ruler.ruler
+ruler_score_group = _ruler.ruler_score_group
 
 __all__ = ["ruler", "ruler_score_group"]


Kovbo:
Let’s keep it as is for now, and reopen it if we have more reasons to fix it.
DEFAULT_RUBRIC is designed to be an internal rubric. `ruler` can accept a custom rubric.

PR #662, opened 2026-04-27
Fix LocalBackend LoRA registry sync

### LINKED ISSUE #661: LoRA adapters are not properly syncing when using LocalBackend - stale adapters are being used in rollouts
When using LocalBackend (colocate mode), I noticed that the LoRA adapter was not being sync properly. If I killed the run after several steps and resumed from a checkpoint, the reward would climb up abruptly as ART was forced to reload the last checkpoint as the first/base model. Upon inspection, it seems that the adapter is correctly making its way to vLLM, but the OpenAI serving layer is not "seeing" the new adapter. As a result, rollouts (which are obtaining the current model using model.get_inference_name()) are just seeing/using the initial/base adapter (step @ 0) and so the whole training process falls apart (silently) due to the stale inference.

This seems to be a regression that may have affected multiple versions, because I do remember this working properly in last year's versions.

While I don't have the means to properly submit a PR at the moment, I wanted to share one possible solution here, in case it helps maintainers:

The problem arises from unsloth/service.py, where at the end of _train_shared() it adds the new adapter to vLLM, but it forgets to also register it within _openai_serving_models. A quick fix would be to add the following snippet right after the llm.add_lora() code block:

```
lora_request = LoRARequest(
	lora_name=f"{self.model_name}@{new_step}",
	lora_int_id=self._next_lora_id(),
	lora_path=checkpoint_dir,
)
added = await llm.add_lora(lora_request)
if not added:
	raise RuntimeError(f"Failed to add LoRA adapter for step {new_step} at {checkpoint_dir}")

# -- Patch here:
import art.vllm.server as _vllm_server_mod
serving_models = _vllm_server_mod._openai_serving_models
if serving_models is not None:
	serving_models.lora_requests[lora_name] = lora_request
	logger.info(
		"Registered '%s' in OpenAI serving models registry", lora_name
	)
else:
	logger.warning(
		"_openai_serving_models is None — LoRA loaded into vLLM "
		"but NOT registered in the OpenAI serving layer. Inference requests "
		"may still use the previous adapter."
	)
# --

self._latest_step = new_step
```

Notes:
- Affected versions: I believe multiple versions are affected, but I personally discovered this with version 0.5.16.
- I acknowledge that this may not be the best place to fix it, but I've verified that ART works correctly after patching this.
- I don't know if train_sft is sensitive to this, but if so, it could be patched in exactly the same way I believe.

@bradhilton this is what we briefly discussed on Discord.

I hope it helps.

bradhilton:
## Summary
Closes #661.

Make LoRA registration update the vLLM OpenAI serving-model registry explicitly after successful shared-mode `add_lora` calls. This keeps `model.get_inference_name()` targets visible to the OpenAI-compatible layer after LocalBackend checkpoint advances.

## Changes
- Added a shared `register_lora_request()` helper for vLLM OpenAI serving-model state.
- Routed Unsloth shared RL, SFT, and skipped-step LoRA registration through one helper that registers the same `LoRARequest` with vLLM and the OpenAI serving registry.
- Applied the same explicit registry update to Megatron shared LoRA aliases.
- Added a unit contract test for the serving-model registry helper.

## Test plan
- `uv run prek run --all-files`
- SkyPilot H200 cluster: `.venv/bin/python -m pytest --tb=short tests/unit/test_vllm_patches_contract.py` (4 passed)
- SkyPilot H200 cluster: `.venv/bin/python -m pytest --tb=short tests/unit/test_multi_checkpoint_inference.py::TestUnslothServiceMaxLoras` (2 passed)
- SkyPilot H200 cluster: `ART_TEST_GPU_MEMORY_UTILIZATION=0.2 ART_TEST_MIN_FREE_GPU_GIB=5 .venv/bin/python -m pytest -s --tb=short tests/integration/test_vllm_contract.py` (1 passed)

Note: local vLLM tests skip on macOS because `vllm` is not installed. The SkyPilot setup command reached a known optional Megatron `deep-ep` build failure due missing `infiniband/mlx5dv.h`, but the backend/vLLM environment was installed enough to run the focused vLLM tests directly from `.venv`.

### DIFF
--- a/src/art/megatron/service.py
+++ b/src/art/megatron/service.py
@@ -31,7 +31,7 @@ from ..utils.convert_moe_lora import convert_checkpoint_if_needed
 from ..utils.get_model_step import get_step_from_dir
 from ..utils.network import find_free_tcp_port
 from ..utils.output_dirs import get_step_checkpoint_dir
-from ..vllm import get_llm, openai_server_task, run_on_workers
+from ..vllm import get_llm, openai_server_task, register_lora_request, run_on_workers
 from .client import create_megatron_job_paths, stream_megatron_job, write_megatron_job
 from .jobs import (
     MegatronMergedTrainJob,
@@ -478,15 +478,15 @@ class MegatronService:
     async def _add_lora_aliases(
         self, llm: AsyncLLM, step: int, checkpoint_dir: str
     ) -> None:
-        added = await llm.add_lora(
-            LoRARequest(
-                lora_name=f"{self.model_name}@{step}",
-                lora_int_id=self._next_lora_id(),
-                lora_path=checkpoint_dir,
-            )
+        lora_request = LoRARequest(
+            lora_name=f"{self.model_name}@{step}",
+            lora_int_id=self._next_lora_id(),
+            lora_path=checkpoint_dir,
         )
+        added = await llm.add_lora(lora_request)
         if not added:
             raise RuntimeError(f"Failed to add LoRA adapter for step {step}")
+        register_lora_request(lora_request)
         self._latest_step = step
 
     async def register_lora_for_step(self, step: int, checkpoint_dir: str) -> None:
--- a/src/art/unsloth/service.py
+++ b/src/art/unsloth/service.py
@@ -26,7 +26,13 @@ from ..utils.convert_moe_lora import convert_checkpoint_if_needed
 from ..utils.get_model_step import get_step_from_dir
 from ..utils.network import find_free_tcp_port
 from ..utils.output_dirs import get_step_checkpoint_dir
-from ..vllm import get_llm, get_worker, openai_server_task, run_on_workers
+from ..vllm import (
+    get_llm,
+    get_worker,
+    openai_server_task,
+    register_lora_request,
+    run_on_workers,
+)
 from .train import (
     UnslothTrainContext,
     create_unsloth_train_context,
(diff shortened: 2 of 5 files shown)


bradhilton:
Blocked on reproduction. I strengthened the vLLM contract test to force a real trainable shared LocalBackend step and then ran it against the PR parent commit (79cb587c) on a SkyPilot H200 cluster with the current backend/vLLM environment. It passed: the new step adapter was present in /v1/models and chat completions with logprobs worked for both @0 and the trained @1 adapter. I also reran the same parent-commit test with LocalBackend(in_process=True), and that passed as well. A v0.5.16-tag check was inconclusive because that tag fails earlier in tokenization with the current dependency environment. Marking this PR draft until we have a clean failing reproduction on current main or a narrowed environment-specific failure.

bradhilton:
Follow-up v0.5.16 repro attempt: created a fresh SkyPilot H200 cluster, checked out tag v0.5.16 in /tmp/art-v0516, and synced that tag's own backend env (transformers 4.57.3, unsloth 2025.12.9, vllm 0.15.1). The old tag needed a temporary tokenization compatibility shim for apply_chat_template returning BatchEncoding before the repro could reach training. With that shim, a focused shared LocalBackend test ran 3 trainable GRPO steps. Each step used model.get_inference_name() for the rollout/current adapter, trained one step, then verified chat completions with logprobs on the new latest name and confirmed the latest adapter appeared in /v1/models. Result: still no clean LoRA-registration/stale-adapter reproduction on v0.5.16.

gmanlan:
I don't know if the versions of the underlying dependencies matter (I would think it doesn't unless openAI or vLLM are doing something automagical WRT adapters registry), but just in case, these are my dep. versions from the environment where I can reproduce this issue:

openai==2.32.0
vllm==0.15.1+cu130
transformers==4.57.3
trl==0.20.0
triton==3.5.0
torch==2.9.0+cu130
unsloth==2025.12.9
unsloth-zoo==2025.12.7
bitsandbytes==0.49.2
peft==0.19.1

Also, in one of the runs (can't remember if before or after monkey patching some stuff to fix this) I noticed that the adapter name (as obtained from model.get_inference_name()) was cycling through the first 3 steps (so @ 1, @ 2, @ 3, then for post-train step 4,5,6 it showed again @ 1, @ 2, @ 3 and so on).

PR #663, opened 2026-04-27
Fix Unsloth autocast dtype for bf16 models

vivekkalyan:
## Summary
PipelineTrainer + LocalBackend can load bf16 base models like Llama 3.1 in 16-bit mode while `ACCELERATE_MIXED_PRECISION` is unset. We were defaulting the Unsloth logprob path to fp16 in that case, which can hit `Half` vs `BFloat16` matmul mismatches.

This keeps explicit mixed-precision settings authoritative, but when the env var is unset, infers the autocast dtype from the loaded model parameters. Unknown mixed-precision/model dtype states fail early instead of silently falling through.

## Validation
- `.venv/bin/pytest tests/unit/test_unsloth_autocast_dtype.py -q`
- Sky 2x H200 smoke: Llama 3.1 + PipelineTrainer + dedicated LocalBackend, `load_in_4bit=False`, `load_in_16bit=True`, `ACCELERATE_MIXED_PRECISION` unset; reached step 2 and reloaded adapters for steps 1 and 2.

### DIFF
--- a/src/art/unsloth/train.py
+++ b/src/art/unsloth/train.py
@@ -313,7 +313,7 @@ def _canonicalize_upstream_metrics(metrics: dict[str, float]) -> dict[str, float
     }
 
 
-def _get_dtype_for_autocasting(trainer: "GRPOTrainer") -> torch.dtype:
+def _get_dtype_for_autocasting(model: torch.nn.Module) -> torch.dtype:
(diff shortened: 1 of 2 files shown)


arcticfly (approved):
Thanks @vivekkalyan!

Issue #668, opened 2026-04-30
Monitor and server tasks are never cancelled on shutdown, causing non-zero exit after training completes


Fiona-Waters:
Description

Two fire-and-forget asyncio tasks are never cancelled during backend shutdown:

1. _monitor_openai_server in LocalBackend (src/art/local/backend.py, line ~491) — created via asyncio.create_task() but the reference is never stored. The monitor polls vLLM health every 30s. After vLLM shuts down, it hits 3 consecutive ConnectionRefusedError failures and re-raises, producing a "Task exception was never retrieved" error and a non-zero process exit.
2. openai_server_task in UnslothService (src/art/unsloth/service.py, line ~516) — in shared mode, the returned asyncio task (the uvicorn server) is discarded. It continues running as an orphaned task after shutdown.

Impact

In containerized environments (Kubeflow TrainJob, KubeRay), the non-zero exit causes the pod to fail/restart even though training completed successfully — adapter checkpoints, metrics, and results are all persisted to disk before the error occurs.

Steps to reproduce

  1. Run a LocalBackend training loop in shared mode (in_process=True) inside a container
  2. Let training complete normally
  3. Observe ConnectionRefusedError in the asyncio task exception output
  4. Process exits non-zero

Error output
```
  Traceback (most recent call last):
    File "/usr/lib64/python3.12/asyncio/tasks.py", line 314, in __step_run_and_handle_result
      result = coro.send(None)
               ^^^^^^^^^^^^^^^
    File "/opt/app-root/lib64/python3.12/site-packages/art/local/backend.py", line 436, in _monitor_openai_server
      async with session.get(
                 ^^^^^^^^^^^^
    File "/opt/app-root/lib64/python3.12/site-packages/aiohttp/client.py", line 1521, in __aenter__
      self._resp: _RetType = await self._coro
                             ^^^^^^^^^^^^^^^^
    File "/opt/app-root/lib64/python3.12/site-packages/aiohttp/client.py", line 788, in _request
      resp = await handler(req)
             ^^^^^^^^^^^^^^^^^^
    File "/opt/app-root/lib64/python3.12/site-packages/aiohttp/client.py", line 742, in _connect_and_send_request
      conn = await self._connector.connect(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/opt/app-root/lib64/python3.12/site-packages/aiohttp/connector.py", line 672, in connect
      proto = await self._create_connection(req, traces, timeout)
[...]

PR #669, opened 2026-04-30
Cancel monitor and server tasks on backend shutdown

Fiona-Waters:
## Summary
- Store `_monitor_openai_server` task references in `LocalBackend` and cancel them during `close()`/`_close()`
- Store `openai_server_task` reference in `UnslothService` and cancel it during `aclose()`/`close()`
- Iterate over `list()` copies to avoid `RuntimeError` from dict mutation when done_callbacks fire during await
- Add unit test verifying `close()` cancels monitor tasks

Fixes #https://github.com/OpenPipe/ART/issues/668

## Test plan
- [ ] Existing `test_local_backend_monitor.py` test passes
- [ ] New `test_close_cancels_monitor_tasks` test passes
- [ ] Run `LocalBackend` training in shared mode inside a container — pod should exit cleanly after training completes

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/art/local/backend.py
+++ b/src/art/local/backend.py
@@ -105,6 +105,7 @@ class LocalBackend(Backend):
 
         # Other initialization
         self._services: dict[str, ModelService] = {}
+        self._monitor_tasks: dict[str, asyncio.Task[None]] = {}
         self._tokenizers: dict[str, PreTrainedTokenizerBase] = {}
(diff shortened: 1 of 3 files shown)


Kovbo:
Good catch, thank you @Fiona-Waters!

Kovbo (approved):

PR #673, opened 2026-05-08
fix: pass **kwargs through to ChatOpenAI in init_chat_model

### LINKED ISSUE #474: `init_chat_model` always uses `ChatOpenAI`, ignores args, and still calls OpenAI / hits 10-minute timeout with Ollama
**Description**

I’m using the ART LangGraph integration and the `init_chat_model` function with a local **Ollama** server as the model backend for my agent. The relevant code in my rollout function looks like this:

```python
chat_model = init_chat_model(model.name, temperature=1.0)
react_agent = create_react_agent(chat_model, tools)
```

I also set up the `art.Model` to point to my local Ollama server for inference (base URL + any key if needed). However, looking at the implementation of `init_chat_model`:

```python
def init_chat_model(...):
    config = CURRENT_CONFIG.get()
    return LoggingLLM(
        ChatOpenAI(
            base_url=config["base_url"],
            api_key=config["api_key"],
            model=config["model"],
            temperature=1.0,
        ),
        config["logger"],
    )
```

Several issues arise:

1. **Arguments are ignored**

   * The `model`, `model_provider`, `configurable_fields`, `config_prefix`, and `**kwargs` are not used at all.
   * This is surprising because I call `init_chat_model(model.name, temperature=1.0)`, but that `model.name` and `temperature` are ignored.
   * Everything is driven purely by the `CURRENT_CONFIG` contextvar.

2. **Hard binding to `ChatOpenAI`**

   * `init_chat_model` *always* creates a `ChatOpenAI` instance, even if I’m not using OpenAI’s API.
   * When using a local Ollama server, it’s still bound to `ChatOpenAI` rather than accepting a generic LangChain chat model (e.g. `ChatOllama` or `ChatNVIDIA`).

3. **Unexpected OpenAI calls / tight OpenAI coupling**

   * Even after pointing the inference base URL to my local Ollama, I still see attempts to call OpenAI endpoints somewhere in the pipeline (especially for judging / RULER).

   * When there’s a runtime error or long inference, the hardcoded **10-minute timeout** in `LoggingLLM.ainvoke` gets hit:

     ```python
     result = await asyncio.wait_for(
         self.llm.ainvoke(input, config=config), timeout=10 * 60
     )
     ```

   * This causes the agent to fail with a timeout, even though:

     * I’m using a local Ollama server.
     * I would like to configure a different timeout for long rollouts.

**What I expect**

* `init_chat_model` should either:

  * **Use its arguments** (`model`, `model_provider`, etc.) and/or accept an explicit `ChatModel` instance; **or**
  * Have a clearer contract that it depends entirely on `CURRENT_CONFIG` and is OpenAI-only.

* It should be possible to plug in:

  * Local Ollama (`ChatOllama` or a Litellm-based wrapper)
  * Other OpenAI-compatible endpoints
  * Without silently falling back to OpenAI-specific assumptions.

* The 10-minute timeout should either be:

  * Configurable; or
  * Documented with guidance on how to override it.

**What actually happens**

* Even after configuring a local Ollama backend and setting inference URLs, ART:

  * Uses `ChatOpenAI` internally.
  * Still exhibits OpenAI-specific behavior.
  * Hits the hard-coded 10-minute timeout during agent inference / error cases.

**Request**

* Please make `init_chat_model` provider-agnostic, or introduce a way to:

  * Pass in a custom `ChatModel` (e.g. `ChatOllama`, `ChatNVIDIA`).
  * Configure the timeout instead of hardcoding `10 * 60`.
* Alternatively, document the intended usage pattern if this function is meant strictly for OpenAI-style backends.

dashitongzhi:
## Problem

`init_chat_model` silently ignores the `model` parameter and all `**kwargs` (including `temperature`, `timeout`, etc.), as reported in #474. Users calling:

```python
chat_model = init_chat_model(model.name, temperature=0.5)
```

find that the model name and temperature are completely ignored — everything is driven solely by the `CURRENT_CONFIG` context variable.

## Changes

- **`init_chat_model`**: Accept `model` as `str | None` (was `Literal[None]`). When a model name is provided, use it; otherwise fall back to `CURRENT_CONFIG`. Build `ChatOpenAI` kwargs with config defaults, then overlay user-provided `**kwargs` so callers can override `temperature`, `max_tokens`, etc.
- **`LoggingLLM`**: Add configurable `timeout` parameter (default: 10 minutes, matching previous hardcoded value). Propagate timeout through `with_structured_output()` and `bind_tools()` so it survives method chains.

## Testing

```python
# Before: temperature=0.5 was silently ignored
init_chat_model("gpt-4", temperature=0.5)

# After: temperature=0.5 is passed to ChatOpenAI
# model name "gpt-4" overrides CURRENT_CONFIG model
```

Fixes #474

### DIFF
--- a/src/art/langgraph/llm_wrapper.py
+++ b/src/art/langgraph/llm_wrapper.py
@@ -108,7 +108,7 @@ def wrap_rollout(model, fn):
 
 
 def init_chat_model(
-    model: Literal[None] = None,
+    model: str | None = None,
     *,
     model_provider: str | None = None,
     configurable_fields: Literal[None] = None,
@@ -116,23 +116,35 @@ def init_chat_model(
     **kwargs: Any,
 ):
     config = CURRENT_CONFIG.get()
+    timeout = kwargs.pop("timeout", 10 * 60)
+    chat_model_kwargs: dict[str, Any] = {
+        "base_url": config["base_url"],
+        "api_key": config["api_key"],
+        "model": model or config["model"],
+        "temperature": 1.0,
+    }
+    chat_model_kwargs.update(kwargs)
     return LoggingLLM(
-        ChatOpenAI(
-            base_url=config["base_url"],  # ty:ignore[unknown-argument]
-            api_key=config["api_key"],  # ty:ignore[unknown-argument]
-            model=config["model"],  # ty:ignore[unknown-argument]
-            temperature=1.0,
-        ),
+        ChatOpenAI(**chat_model_kwargs),
         config["logger"],
+        timeout=timeout,
     )
 
 
 class LoggingLLM(Runnable):
-    def __init__(self, llm, logger, structured_output=None, tools=None):
+    def __init__(
+        self,
+        llm,
+        logger,
+        structured_output=None,
+        tools=None,
+        timeout: float = 10 * 60,
+    ):
         self.llm = llm
         self.logger = logger
         self.structured_output = structured_output
         self.tools = [convert_to_openai_tool(t) for t in tools] if tools else None
+        self.timeout = timeout
 
     def _log(self, completion_id, input, output):
         if self.logger:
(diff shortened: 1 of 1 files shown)


Kovbo:
Thanks, fixing [here](https://github.com/OpenPipe/ART/pull/713)

PR #681, opened 2026-05-17
Preserve RNG state in dataset iteration

pragnyanramtha:
## Summary
- avoid resetting Python global random state while shuffling dataset epochs
- keep deterministic epoch ordering by using a local RNG seeded per epoch
- add focused tests for deterministic iteration and global RNG preservation

## Validation
- uv run pytest tests/unit/test_iterate_dataset.py
- uv run ruff format --check src/art/utils/iterate_dataset.py tests/unit/test_iterate_dataset.py
- uv run ruff check src/art/utils/iterate_dataset.py tests/unit/test_iterate_dataset.py


Kovbo:
Good catch, thank you, @pragnyanramtha!

PR #702, opened 2026-06-01
PipelineTrainer KL step lag

Kovbo:
## Summary

- Rebase the KL step-lag behavior from #625 onto the current main branch without changing the contract.
  - kl_penalty_coef=0: KL disabled
  - kl_penalty_coef>0 and kl_penalty_step_lag=None: use checkpoint 0
  - kl_penalty_coef>0 and kl_penalty_step_lag>=1: use max(0, current_step - lag)
  - kl_penalty_step_lag=0: invalid


src/art/megatron/train.py
```diff
@@ -481,6 +483,43 @@ def run_megatron_rl_job(
             job.config.grad_accumulation_sequences
         )
         num_steps = math.ceil(num_sequences / global_grad_accumulation_sequences)
+        ref_adapter_path = cast(dev.TrainConfig, job.experimental_config).get(
```
FurtherAI: The call to the megatron model seems to be roughly duplicated with the version in `run_training_step`. In both locations, we want to compute the megatron logprobs. The reference runs without grad but that could be under a context manager. So, we should see if we can abstract those to a shared, clean function. 

Also, we could separate the whole reference adapter part, 489-522 into a function so `run_megatron_rl_job` stays somewhat clean to read.

src/art/megatron/train.py
```diff
@@ -481,6 +483,43 @@ def run_megatron_rl_job(
...
+            "kl_ref_adapter_path"
+        )
+        if job.config.kl_penalty_coef > 0.0 and ref_adapter_path is not None:
```
FurtherAI: Also, do you want it to just print a warning if the KL penalty is on but the ref path is missing? Should it ever be in that state?

FurtherAI:
@codex does this handle adapter leases correctly so the reference step doesn't get cleaned up before it is used?

src/art/megatron/train.py
Kovbo: In the PipelineTrainer path, KL-on always produces a `kl_penalty_reference_step`, and LocalBackend.train() resolves that into `kl_ref_adapter_path` before the Megatron job is written. So Megatron should not see `kl_penalty_coef > 0` with a missing ref path.

angkywilliam (approved):

PR #714, opened 2026-06-04
chore: Improve Megatron runtime typing

bradhilton:
## Summary
- Tighten Megatron runtime/provider/helper type hints while skipping Pydantic runtime validation for arbitrary Megatron objects.
- Alias `ModelChunk` to `MegatronModule` and keep validation helpers for Megatron API boundaries.
- Forward TE `backward_dw()` through LoRA wrappers so delayed weight-gradient hooks remain available after wrapping.

## Test plan
- `uv run prek run --all-files`
- `uv run pytest --tb=short tests/integration/test_lora_quack_cutover.py tests/integration/megatron/lora/test_merged_weight_export.py tests/unit/test_pipeline_trainer_local_backend.py::test_load_adapter_into_model_reloads_optimizer_when_provided tests/unit/test_model_support_import_boundary.py tests/unit/test_moe_routing_replay.py tests/unit/test_moe_routing_real_path.py tests/integration/megatron/model_support/test_hf_parity_invariants.py tests/integration/megatron/model_support/test_compile_flags.py tests/integration/megatron/model_support/test_workflow.py tests/integration/megatron/model_support/test_packed_position_ids.py tests/integration/megatron/model_support/test_provider_support.py`
- Minimal Megatron yes/no/maybe smoke via `dev/yes-no-maybe-megatron.py`
- Synthetic Megatron gradient smoke with one trainable trajectory group

### DIFF
--- a/src/art/megatron/lora.py
+++ b/src/art/megatron/lora.py
@@ -229,7 +229,9 @@ def _set_lora_shard_strategy_metadata(
 
 def _exported_shard_dim(param: torch.nn.Parameter) -> int:
     assert hasattr(param, "lora_tp_shard_dim")
-    axis = _normalize_axis(getattr(param, "lora_tp_shard_dim"), param.ndim)  # ty: ignore[unresolved-attribute]
+    axis = _normalize_axis(
+        getattr(param, "lora_tp_shard_dim"), param.ndim
+    )  # ty: ignore[unresolved-attribute]
     # LoRA exports always serialize a 2D tensor:
     # - non-expert params export `param.T`
     # - expert params export `param[expert].T`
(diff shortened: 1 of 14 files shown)


src/art/megatron/lora.py
```diff
@@ -276,9 +293,9 @@ def num_local_experts(self) -> int:
         return self.A_T.shape[0] if self.A_T.ndim == 3 else 1
 
     def _broadcast_if_replicated(self, param: torch.nn.Parameter) -> None:
-        if not param.lora_tp_replicated:  # ty: ignore[unresolved-attribute]
+        if not getattr(param, "lora_tp_replicated"):
```
FurtherAI: This kind of thing (not being defensive about attributes) was deliberate on my part. I would like to minimize the amount of defensive/fallback behavior that Codex inserts and I don't put that much effort into it, but this is one thing. I don't see a reason to be defensive here

bradhilton: it's not really defensive, both will raise if the attribute is not present, just the second made my type checker happy. i will use a different approach though.

FurtherAI (approved):

PR #721, opened 2026-06-05
Fix routing replay split sizes for attention

vivekkalyan:
## Summary
- preserve the full attention token layout when recording routing replay token UID sets
- keep the compacted GDN token UID layout only for GDN replay
- add a unit test that covers the differing attention-vs-GDN replay shapes

## Why
The Megatron routing replay path could fail with `split_with_sizes expects split_sizes to sum exactly ...` when attention replay used compacted GDN token UID sets. Attention needs the original flattened token layout, while GDN uses the compact routed-token layout.

## Validation
- `uv run --with torch --with safetensors --with megatron-core==0.17.0 --with transformers==5.2.0 --group dev pytest tests/unit/test_moe_routing_replay.py tests/unit/test_dedicated_config.py`
- Sky 2x H200 Bonnie Megatron repro against this fix completed 1 training step without the split-size crash
- Stacked LoRA PR smoke run against this branch also completed 1 Megatron training step successfully

### DIFF
--- a/src/art/megatron/training/trace.py
+++ b/src/art/megatron/training/trace.py
@@ -98,19 +98,17 @@ def _routing_replay_token_uid_sets(
     *,
     attention_state: Any | None,
 ) -> dict[str, torch.Tensor | None]:
+    attention_token_uids = flatten_local_token_uids(token_uids)
     plan = getattr(attention_state, "gdn_execution_plan", None)
     if plan is not None:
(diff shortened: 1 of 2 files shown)


FurtherAI:
Looks good, seems like a misconception from Codex and not caught by tests because they fill the packed seq and don't have padding.

PR #676, opened 2026-05-11
Support cross-entity checkpoint forks

poofeth:
Fixes #649.

## Summary
- Add an optional `from_entity` parameter to `ServerlessBackend._experimental_fork_checkpoint`.
- Use the explicit source entity when building the W&B artifact collection path, falling back to the destination model entity and then W&B default entity.
- Add unit coverage for explicit source entity, fallback behavior, and missing-entity validation.

## Validation
- `uv run pytest tests/unit/test_serverless_fork_checkpoint.py -q`
- `uv run ruff check src/art/serverless/backend.py tests/unit/test_serverless_fork_checkpoint.py`
- `uv run ruff format --check src/art/serverless/backend.py tests/unit/test_serverless_fork_checkpoint.py`
- `git diff --check`

Note: this is a regular issue fix, not an Algora bounty claim.

PR #741, opened 2026-06-27
feat(serverless): support forking checkpoints from a different W&B entity

### LINKED ISSUE #649: Add from_entity parameter to _experimental_fork_checkpoint
## Problem

`_experimental_fork_checkpoint` in `ServerlessBackend` constructs the source artifact path using the **destination model's entity** (line ~89):

```python
from_entity = model.entity or api.default_entity
collection_path = f"{from_entity}/{from_project}/{from_model}"
```

This means you can't fork from a checkpoint in a different W&B entity. For example, forking from `willow-voice/willow_normal/kl-000-1` into `wb-training/willow_normal/my-new-run` fails because it looks for the artifact under `wb-training`.

## Proposed Fix

Add an optional `from_entity` parameter to `_experimental_fork_checkpoint`:

```python
async def _experimental_fork_checkpoint(
    self,
    model: Model,
    from_model: str,
    from_project: str | None = None,
    from_entity: str | None = None,  # NEW
    from_s3_bucket: str | None = None,
    not_after_step: int | None = None,
    verbose: bool = False,
    prefix: str | None = None,
) -> None:
```

And use it when constructing the artifact path:

```python
from_entity = from_entity or model.entity or api.default_entity
```

## Current Workaround

Download the artifact from the source entity and re-upload it to the destination entity before calling fork. This works but doubles the artifact storage and adds an unnecessary copy step.

AnnaSuSu:
## What

`ServerlessBackend._experimental_fork_checkpoint` builds the source W&B artifact path from the **destination** model's entity:

```python
from_entity = model.entity or api.default_entity
collection_path = f"{from_entity}/{from_project}/{from_model}"
```

So a checkpoint can only be forked within the same entity — forking e.g. `willow-voice/willow_normal/kl-000-1` into `wb-training/willow_normal/my-new-run` fails because it looks for the artifact under `wb-training`. The only workaround is to download from the source entity and re-upload to the destination, doubling storage.

Closes #649.

## Fix

Add an optional `from_entity` parameter to `_experimental_fork_checkpoint` and resolve the source entity as `from_entity → model.entity → api.default_entity`. The resolution + path building move into a small pure helper, `_wandb_checkpoint_collection_path`, which also raises a clear `ValueError` when no entity can be determined (previously this produced a `"None/…"` path).

```python
await backend._experimental_fork_checkpoint(
    model,                       # destination, e.g. wb-training/...
    from_model="kl-000-1",
    from_project="willow_normal",
    from_entity="willow-voice",  # NEW: source entity
)
```

This re-implements the approach from #676 (validated there but voluntarily closed by its author); credit to @poofeth for the original.

## Tests

New `tests/unit/test_serverless_fork_checkpoint.py` — no GPU/backend deps, runs on the base install:

- `_wandb_checkpoint_collection_path` resolution: explicit `from_entity` wins → falls back to model entity → falls back to default entity → raises when none.
- An async test drives `_experimental_fork_checkpoint` through the W&B branch with a fake `wandb.Api` and asserts the artifact query uses the explicit source entity (`("lora", "src-entity/src-project/src-model")`).

```
tests/unit/test_serverless_fork_checkpoint.py ..... [5 passed]
existing serverless unit tests: 12 passed (no regressions)
```

### DIFF
--- a/src/art/serverless/backend.py
+++ b/src/art/serverless/backend.py
@@ -48,6 +48,26 @@ def _extract_step_from_wandb_artifact(artifact: "Artifact") -> int | None:
     return None
 
 
+def _wandb_checkpoint_collection_path(
+    *,
+    from_model: str,
+    from_project: str,
+    model_entity: str | None,
+    default_entity: str | None,
+    from_entity: str | None = None,
+) -> str:
+    """Build the W&B artifact collection path for a source checkpoint.
+
+    Resolves the entity from the explicit ``from_entity`` first, then the
+    destination model's entity, then the W&B default entity, so a checkpoint can
+    be forked from an entity other than the destination's.
+    """
+    resolved_entity = from_entity or model_entity or default_entity
+    if resolved_entity is None:
+        raise ValueError("A W&B entity is required to locate the source checkpoint")
+    return f"{resolved_entity}/{from_project}/{from_model}"
+
+
 _UPSTREAM_TRAIN_METRIC_KEYS = {
     "reward": "reward",
     "reward_std_dev": "reward_std_dev",
@@ -881,6 +901,7 @@ class ServerlessBackend(Backend):
         model: "Model",
         from_model: str,
         from_project: str | None = None,
+        from_entity: str | None = None,
         from_s3_bucket: str | None = None,
         not_after_step: int | None = None,
         verbose: bool = False,
@@ -899,6 +920,10 @@ class ServerlessBackend(Backend):
             model: The destination model to fork to.
             from_model: The name of the source model to fork from.
             from_project: The project of the source model. Defaults to model.project.
+            from_entity: The W&B entity of the source model. Defaults to
+                model.entity, then the W&B API's default entity. Set this to fork
+                from a checkpoint that lives in a different entity than the
+                destination model.
             from_s3_bucket: Optional S3 bucket to pull the checkpoint from.
             not_after_step: If provided, uses the latest checkpoint <= this step.
             verbose: Whether to print verbose output.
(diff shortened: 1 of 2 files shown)


AnnaSuSu:
any chance of a look here? adds support for forking checkpoints from a different W&B entity. it's mergeable and ready, glad to adjust if you'd want it done differently.

AnnaSuSu:
rebased onto current main (88fad83) — the conflict is resolved and the PR is mergeable again.

on the conflict itself: #739 swapped the serverless backend's `import wandb` / `wandb.Api(...)` for the `..utils.wandb_sdk` lazy-import shim. that landed on the line directly above the one this PR removes (`from_entity = model.entity or api.default_entity`, which becomes a parameter here), so the two changes were textually adjacent but not semantically at odds. the resolution takes `wandb_sdk.api(...)` and keeps `from_entity` as the new parameter — `wandb_sdk.api()` returns the same `wandb.apis.public.Api`, so `.default_entity` and `.artifacts()` behave as before.

the rebase did require one real change. the test used to stub W&B by injecting a fake module into `sys.modules`, which no longer works now that the shim does `from wandb.apis.public import Api` (it raises `ModuleNotFoundError: No module named 'wandb.apis'; 'wandb' is not a package`). it now patches at the seam #739 established, matching `tests/unit/test_serverless_pipeline_trainer_compat.py`:

```python
monkeypatch.setattr("art.serverless.backend.wandb_sdk.api", FakeApi)
```

verified locally on the rebased branch:

- `uv run pytest tests/unit/test_serverless_fork_checkpoint.py` → 5 passed. reverting the fix reddens the end-to-end test with exactly the bug in #649 (it queries `dst-entity/src-project/src-model` instead of `src-entity/...`).
- full `tests/unit` failure set is byte-identical to clean main: 13 pre-existing failures + 20 collection errors, all from torch / transformers / tinker / datasets not being installed in my local venv. 182 passed on main, 187 with this PR (the 5 new ones).
- `uv run prek run --all-files`: ruff, ruff format and the uv.lock sync check all pass. the only `ty` diagnostics are pre-existing `unresolved-import`s under `src/art/megatron/**` (torch/tilelang absent locally) — none in either file this PR touches.

@bradhilton the workflows on this head are sitting at `action_required` — fork PRs need a maintainer to click *Approve and run*, so no checks are displayed. could you approve the run when you get a moment?

PR #750, opened 2026-07-11
Upgrade W&B to 0.28

bradhilton:
## What changed

- upgrade the backend W&B pin from 0.25.0 to 0.28.0
- pin Click 8.2.0 through ART's existing uv override mechanism because W&B 0.28 requires Click >=8.2 while SkyPilot 0.11.1 declares Click <8.2
- refresh only the affected lock entries

## Why

Caladan is moving to W&B 0.28.0 and ART should exercise the same artifact and run behavior. The Click override is the smallest resolvable compatibility bridge; both W&B and SkyPilot CLIs were exercised under the combined environment.

## Validation

- full `prek run --all-files`: passed (Ruff, format, ty, lock check)
- exact backend/Tinker CI unit command: 276 passed, 3 skipped
- `wandb --help`: passed
- `sky --help`: passed
- imports verified: W&B 0.28.0, SkyPilot 0.11.1, Click 8.2.0

A raw repository-wide pytest invocation is not a supported single environment because the Megatron, backend, and Tinker extras are intentionally split; the applicable CI matrix command above passed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ backend = [
     "accelerate==1.7.0",
     "awscli>=1.38.1",
     "setuptools>=78.1.0",
-    "wandb==0.25.0",
+    "wandb==0.28.0",
     "transformers==5.2.0",
     "duckdb>=1.0.0",
     "pyarrow>=15.0.0",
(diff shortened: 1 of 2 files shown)


bradhilton:
Superseded by #754.

PR #778, opened 2026-07-27
fix: retry transient vLLM adapter reload failures

### LINKED ISSUE #678: Dedicated LocalBackend vLLM server can become unreachable during LoRA adapter reload
## Summary

In LocalBackend dedicated mode with PipelineTrainer, some runs fail or become unhealthy immediately after an early training step when ART tries to reload the newly saved LoRA adapter into the dedicated vLLM server.

The observed failure is an `httpx.ConnectError: All connection attempts failed` from `UnslothService._reload_adapter(...)`. After this starts happening, rollout workers also repeatedly fail with generic `Connection error` while calling `/chat/completions`.

This was seen after upgrading ART to main commit `48b2e5f6c384a62b44f34e1472e5fb1eeaa3474a`.

## Setup

- ART commit: `48b2e5f6c384a62b44f34e1472e5fb1eeaa3474a`
- Backend: `LocalBackend`
- Trainer: `PipelineTrainer`
- Mode: dedicated trainer/inference GPUs
- Hardware: 1 node, 2x H200; one GPU for trainer and one GPU for inference
- Base model class: Llama 3.1 8B Instruct
- Training: GRPO/CISPO via Unsloth
- vLLM from the ART lock at this commit (`vllm 0.17.0+art1` in the generated lock)
- Checkpoint saving enabled
- `eval_at_start` enabled

No private data is needed to reproduce the failure mode; this appears to be an infrastructure / adapter reload stability issue.

## Symptoms

One run failed after the first training step with this stack:

```text
Pipeline stage failed: All connection attempts failed
ExceptionGroup: unhandled errors in a TaskGroup
  File ".../art/pipeline_trainer/trainer.py", line 467, in _training_stage
    result = await self.backend.train(...)
  File ".../art/unsloth/service.py", line 628, in _train_dedicated
    await self._reload_adapter(checkpoint_dir, new_step)
  File ".../art/unsloth/service.py", line 455, in _reload_adapter
    response = await client.post(...)
  File ".../httpx/_client.py", line 1859, in post
    return await self.request(...)
httpx.ConnectError: All connection attempts failed
```

Other concurrent runs on the same setup remained marked as running, but stopped making useful training progress and emitted repeated rollout failures like:

```text
Worker N: rollout failed: Connection error.
Retrying request to /chat/completions ...
```

## Expected behavior

If the dedicated vLLM server is temporarily unavailable during adapter reload, ART should either:

1. Wait for the server to become healthy again before resuming rollouts/training, or
2. Restart/recover the dedicated inference server, or
3. Fail with a clearer diagnostic that includes vLLM process health and recent server logs.

A transient adapter reload/server restart should not leave the pipeline in a state where rollout workers keep generating connection errors indefinitely.

## Actual behavior

At least one run hard-failed during `_reload_adapter(...)` after saving a checkpoint. Other runs kept running at the Sky/job level but repeatedly failed rollout calls to the local vLLM endpoint and did not advance training meaningfully.

## Possibly related

This was found while investigating forked LocalBackend/PipelineTrainer runs. There is already an open fork-specific issue here:

- #651

The adapter reload/connectivity failure described in this issue is separate from whether the run started from the intended fork checkpoint. Even after ignoring the fork correctness problem, the dedicated vLLM server becoming unreachable during or after adapter reload appears to be a separate stability issue.

## Suggested debugging additions

It would help if `_reload_adapter(...)` logged or surfaced:

- Target URL/endpoint being called
- Whether the vLLM subprocess is still alive
- Last lines of the dedicated vLLM log file
- Health check result before and after adapter reload
- The checkpoint path and step being reloaded

yaodong-shen:
## Summary

- retry LoRA reloads after connection establishment failures and transient 502/503/504 responses
- check the supervised vLLM child before every attempt so a real process exit stops retries immediately
- report the request URL, process state, runtime log path, checkpoint step/path, and final error when retries are exhausted

Fixes #678

## Retry safety

The request is retried at most four times with exponential backoff. Client errors and read/write failures are not retried because the server may already have applied the non-idempotent adapter load.

## Validation

- uv run --no-sync pytest -q tests/unit/test_vllm_runtime.py — 5 passed
- existing runtime launcher tests — 14 passed; one macOS /tmp versus /private/tmp assertion failed identically on a clean origin/main@3bdbf229 worktree
- prek ruff, ruff-format, ty, and uv-lock-check hooks — passed
- Python 3.12 compile and git diff --check — passed

The complete unit suite requires optional backend and GPU dependencies that are not installable in this macOS environment; repository CI installs those extras in its CUDA runner.

### DIFF
--- a/src/art/unsloth/service.py
+++ b/src/art/unsloth/service.py
@@ -566,9 +566,6 @@ class UnslothService:
 
     async def _reload_adapter(self, checkpoint_path: str, step: int) -> None:
         """Reload LoRA adapter in vLLM subprocess via HTTP."""
-        import httpx
-
-        self._raise_if_child_failed()
         lora_name = f"{self.model_name}@{step}"
         logger.info(
             f"[DEDICATED] _reload_adapter START: lora_name={lora_name} "
@@ -580,14 +577,14 @@ class UnslothService:
         }
         if self.serving_capabilities.inplace_lora_load:
             payload["load_inplace"] = True
-        async with httpx.AsyncClient() as client:
-            response = await client.post(
-                f"{self._vllm_base_url}/v1/load_lora_adapter",
-                json=payload,
-                **self._runtime_request_kwargs(),
-                timeout=60.0,
-            )
-            response.raise_for_status()
+        response = await self._vllm_runtime.post_with_retry(
+            "/v1/load_lora_adapter",
+            json=payload,
+            timeout=60.0,
+            before_attempt=self._raise_if_child_failed,
+            operation="LoRA adapter reload",
+            failure_context=f"step={step}; checkpoint={checkpoint_path}",
+        )
         logger.info(
             f"[DEDICATED] _reload_adapter DONE: lora_name={lora_name} "
             f"status={response.status_code}"
--- a/src/art/vllm_runtime.py
+++ b/src/art/vllm_runtime.py
@@ -3,6 +3,7 @@ from contextlib import contextmanager
 import fcntl
 import hashlib
 import json
+import logging
 import math
 import os
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


yaodong-shen:
The Prek, Package Install, and TrainerRank GPU workflows are waiting for maintainer approval and created no jobs for this external-fork PR. The focused retry tests and the repository Prek checks pass locally. Could a maintainer approve the workflow runs when convenient?

PR #796, opened 2026-08-14
Fix multi-turn reasoning tokenization, LoRA recompile churn, and checkpoint slot refs

bradhilton:
## Context

Companion to OpenPipe/caladan#109. A long-running tau2-bench/Qwen3.5-4B campaign exposed multi-turn reasoning re-render failures, trainer Dynamo recompiles, and checkpoint restore edge cases. This PR contains the ART-side repairs.

> [!IMPORTANT]
> The original soak was not a clean-clone validation: the driver imported ART through a stale editable checkout, while the trainer used the older ART revision pinned by Caladan. That run was useful for diagnosis but is not treated as verification of this final branch. Clean-environment driver/trainer SHA validation is part of the companion Caladan PR's merge gate.

### Preserve prior thinking end to end

- ART-managed trainable and internally configured models default to `preserve_thinking=true`; explicit caller overrides, including `false`, still win. Unmanaged external endpoints are unchanged.
- Legacy Qwen3/Qwen3.5 templates are upgraded only when needed, including hub namespaces and local paths whose final component is a Qwen3/Qwen3.5 model name.
- Explicit server chat templates take precedence and avoid tokenizer/hub access. Expected offline or invalid-tokenizer loads warn and fall back instead of failing startup.
- SFT and trajectory re-rendering share the preservation behavior. The intentional SFT behavior is documented, with multi-turn Qwen regression coverage.

### Tool-call argument normalization

- Shared normalization converts JSON-string tool arguments to mappings only for templates that iterate the tool-call arguments: `arguments | items`, direct `arguments.items()`, or a recognized tool-call argument alias subsequently iterated with `.items()`.
- GLM/MiniMax-style positives and gpt-oss/schema-object negatives prevent the earlier broad `.items()` heuristic from changing unrelated templates.
- Empty no-argument strings normalize to `{}`; invalid non-empty or non-object JSON raises a clear error.

### History reconciliation and alternate rendering

- Speculative chat-template probes are non-fatal; errors from the authoritative render still propagate.
- Marked message bounds are recomputed where possible and invalidated after alternate-render mismatches or failed probes, preventing stale offsets from being reused.
- Canonical marker/probe spans are translated across both same-length and length-changing exact prompt splices, including a canonical-empty prompt.
- Reconciliation keys include hidden reasoning and canonicalize the `reasoning_content` alias, so visibly identical histories with different reasoning cannot borrow one another's sampled-source metadata.
- Dict-valued tokenizer template registries are not forwarded as a concrete chat template.

### Checkpointing and compilation

- Weights-only checkpoints remain valid and lazily recreate optimizer state; LoRA slot refs are preserved across restore failures.
- Dynamic Dynamo parameter shapes are enabled only with transformer compilation. This is intentionally process-wide because PyTorch exposes the setting globally; regression coverage locks down the conditional application.

### Rollout observability

- Tau-bench trajectories record policy, environment, startup, active latency, and prompt/completion token metrics.

## Testing

- Final tokenization/SFT regression set: **181 passed**; two additional end-to-end tests require the optional `tinker` dependency absent from the local environment.
- Fresh Codex and Claude Fable 5 findings were addressed with focused regressions and clean follow-up reviews.
- Ruff formatting/lint and `ty` checks pass for all touched files.
- Megatron compilation regressions: **4 passed**.
- Checkpoint/tau regressions: **106 passed**.
- GitHub Actions is the authoritative clean-environment check for the final head.

### DIFF
--- a/docs/features/additional-histories.mdx
+++ b/docs/features/additional-histories.mdx
@@ -27,6 +27,12 @@ Each trajectory can contain:
 
 Some models, like Qwen 3, use chat templates that remove special tokens (such as `<think>`) from previous turns in multi-turn conversations. This can interfere with training when you want the model to learn from its thinking process across all turns.
 
+ART-managed inference and SFT paths preserve prior-turn thinking by default when
+the model's chat template supports it. You can override that behavior explicitly
+with `chat_template_kwargs={"preserve_thinking": False}`. Additional histories
+remain useful for custom or externally managed templates that do not expose a
+prior-thinking preservation option.
+
 By splitting each turn into a separate history, you can preserve these tokens for training:
 
 ```python
--- a/src/art/local/backend.py
+++ b/src/art/local/backend.py
@@ -16,6 +16,10 @@ from types import TracebackType
 from typing import TYPE_CHECKING, Any, AsyncIterator, Iterable, Literal, cast
 import warnings
 
+from art.utils.chat_template import (
+    chat_template_with_preserved_thinking,
+    configure_preserved_thinking_chat_template,
+)
 from art.utils.lifecycle import (
     PROCESS_SHUTDOWN_TIMEOUT_SECONDS,
     process_shutdown_timeout,
@@ -242,6 +246,14 @@ def _apply_configured_chat_template_server_args(
         chat_template = _model_support_default_chat_template(
             base_model, internal_config
         )
+    if chat_template is None and base_model is not None and base_model.startswith(
+        ("Qwen/Qwen3-", "Qwen/Qwen3.5-", "OpenPipe/Qwen3-")
+    ):
+        tokenizer = AutoTokenizer.from_pretrained(base_model)
+        default = getattr(tokenizer, "chat_template", None)
+        preserved = chat_template_with_preserved_thinking(default)
+        if preserved != default:
+            chat_template = cast(str, preserved)
     if chat_template is None:
         return
     server_args = dict(config_dict.get("server_args", {}))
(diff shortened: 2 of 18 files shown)


bradhilton:
Post-remediation review resolution for final head `e5ceec45`:

- Fresh Codex found structured tool arguments losing sampled provenance, speculative render errors escaping, and overly broad nested-schema detection. These were fixed with focused positive/negative regressions.
- Claude Fable 5 found canonical probe spans could be computed against an exact-prefix-spliced render, plus three small reasoning/tool edge cases. Probe spans now derive from canonical token space and translate into the spliced render; reasoning-only empty token IDs are treated as missing evidence; empty no-argument strings normalize to `{}`; and reasoning aliases use the same slot-selection rule.
- A final fresh Codex pass reproduced the same canonical/spliced mismatch for marker offsets on length-changing prompts. The translation now covers marker bounds, whitespace alternate renders, empty-message bounds, single-part probes, stale-bound correction, and the single-sampled fallback, including a canonical-empty/exact-nonempty prompt.
- Claude also noted that `_artifact_base_model()` can fall back from a versioned artifact lacking `base_model` metadata to the lineage's `:latest` metadata. That behavior predates this PR and is outside the thinking/tokenization/checkpoint remediation; changing artifact lineage semantics here would expand the PR materially, so it is intentionally left for a separate issue.
- Claude's remaining low/info notes are intentionally unchanged: audio-only Chat Completions are outside ART's supported training path; mixed-key tool-call dicts are unreachable through validated histories; generated `{}` for a truly empty argument string is renderer scaffold and should remain unsampled; and simplifying vestigial defensive conditions would add churn without changing behavior. The same-length/length-changing marker-offset coverage gap was closed in the final test-only commit; the empty-message path shares the now-tested translation helper, and constructing a synthetic reconciled multi-turn empty-output source solely to hit that branch would add disproportionate fixture complexity.

Final local tokenization/SFT set: 181 passed, with two optional-`tinker` tests deselected. Touched-file Ruff and `ty` checks pass. GitHub Actions remains the clean-environment authority.

PR #801, opened 2026-08-16
Fix Gemma 4 MoE LoRA tensor layout

Kovbo:
## Summary
- detect PEFT’s transposed Gemma 4 MoE target-parameter LoRA layout
- transpose and swap its factors before vLLM normalization or Megatron import
- cover both export and import paths with a focused regression test

## Validation
- `uvx ruff check src/art/megatron/model_support/handlers/gemma4.py tests/integration/megatron/lora/test_lora_disk_codecs.py`
- `uvx ruff format --check src/art/megatron/model_support/handlers/gemma4.py tests/integration/megatron/lora/test_lora_disk_codecs.py`
- `python3 -m compileall -q src/art/megatron/model_support/handlers/gemma4.py tests/integration/megatron/lora/test_lora_disk_codecs.py`
- production checkpoint smoke test: converted 30 layers / 15,770 internal tensors with expected Gemma 4 MoE shapes

The focused Megatron test cannot collect on macOS because Megatron-Bridge depends on Linux-only `nvidia-resiliency-ext`; CI provides the supported Linux environment.

### DIFF
--- a/src/art/megatron/model_support/handlers/gemma4.py
+++ b/src/art/megatron/model_support/handlers/gemma4.py
@@ -1817,11 +1817,87 @@ def _vllm_moe_config(adapter_config: dict[str, Any]) -> dict[str, Any]:
     return config
 
 
+def _canonicalize_gemma4_peft_moe_target_parameter_layout(
+    tensors: dict[str, torch.Tensor],
+    *,
+    adapter_config: dict[str, Any],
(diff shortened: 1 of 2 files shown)


Kovbo:
Fast-forwarded this validated commit into the base branch for PR #800 so the production dependency remains pinned to the tested SHA. Closing this now-empty stacked PR.

PR #803, opened 2026-08-17
Fix Qwen 3.6 MoE identity LoRA layout

Kovbo:
## Summary

Fix Qwen 3.5/3.6 MoE LoRA conversion so identity adapters created through PEFT can be loaded by Megatron and normalized for vLLM.

## What failed

Qwen/Qwen3.6-35B-A3B training failed before the first optimization step while Megatron loaded the rank-1 identity LoRA:

    gate_up_proj: sharded load shape mismatch, got (256, 1024, 1), expected (256, 2048, 1)

The saved expert factors were transposed and assigned to the wrong side of each LoRA pair. For example, gate/up A was `(256, 1024)` instead of `(256, 2048)`, and the down-projection factors had the analogous mismatch.

## What we tried and why it did not work

We first checked whether serverless training was loading an old or partially generated identity adapter. The failing adapter had just been generated by the identity-LoRA workflow, so it was not stale.

Generating it again with the same ART/PEFT conversion path produced the same layout. A cache refresh alone therefore could not fix the problem; it would only store another malformed adapter.

We then compared the Qwen parameter shapes, PEFT factors, vLLM representation, and Megatron loader expectations. PEFT interprets a 3D target parameter as `[experts, in, out]`, while Qwen stores these expert weights as `[experts, out, in]`. That reverses the A/B dimensions for gate/up and down expert projections.

## What this fixes

- Detect the PEFT expert layout using Qwen's hidden size, MoE intermediate size, and expert count.
- Swap and transpose the affected gate/up and down A/B factors into the canonical layout.
- Apply the normalization before both vLLM conversion and Megatron import so the two paths agree.
- Add a regression test covering the PEFT-style Qwen 3.6 expert layout.

## Validation

- [x] Ruff check and format check.
- [x] Local tensor-layout regression smoke test.
- [x] Generated a fresh Qwen 3.6 35B rank-1 identity LoRA on an H200.
- [x] Loaded that adapter in Megatron without the original shape mismatch.
- [x] Completed a one-batch Qwen 3.6 35B SFT job on the H200.

### DIFF
--- a/src/art/megatron/model_support/handlers/qwen3_5.py
+++ b/src/art/megatron/model_support/handlers/qwen3_5.py
@@ -651,6 +651,83 @@ def _vllm_moe_config(
     return config
 
 
+def _canonicalize_qwen35_peft_moe_target_parameter_layout(
+    tensors: dict[str, torch.Tensor],
+    *,
+    adapter_config: dict[str, Any],
+) -> dict[str, torch.Tensor]:
+    """Transpose PEFT's parameter-LoRA factors into vLLM's 3D MoE layout.
+
+    Qwen3.5 and Qwen3.6 store expert parameters as ``[expert, out, in]``
+    tensors, while PEFT's generic 3D target-parameter wrapper interprets them
+    as ``[expert, in, out]``. The resulting factors already use vLLM's fused
+    expert key names, but every expert delta is transposed. Shape-detect that
+    layout and transpose the factors back before Megatron import or vLLM
+    normalization.
+    """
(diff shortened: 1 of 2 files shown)


Kovbo:
Superseded by #806. The 3D MoE LoRA layout bug is fixed upstream in PEFT 0.19.1; #806 makes the supported PEFT range explicit for both ART backend and Megatron consumers. Serverless Training will invalidate the affected identity-LoRA cache, so the ART-side malformed-checkpoint recovery in this PR is not needed.

PR #826, opened 2026-09-01
Replace TrainerRank policy knobs with an automatic prefix-tree layout planner

bradhilton:
## Summary

Replaces TrainerRank's user-tuned policy knobs with an automatic, data-dependent prefix-tree layout planner. `TrainerRank(runtime)` is now the entire constructor.

**Scope, precisely:** prefix sharing (arbitrary depth, per-subtree share/replay) is a data-dependent planner decision; microbatch width reuses main's adaptive selector, now sharing-aware with a monotone feasibility predicate (a width is feasible iff its memory-minimal full-sharing layout fits; the cost-optimal layout executes when it fits, the memory-minimal one otherwise, and the chosen mode is what gets materialized — measured: the same memory budget admits 2 waves/step where the no-sharing bound forced 4, and the step is 16% faster than that 4-wave baseline; planner pricing runs only inside the band where the cheap no-sharing and full-sharing bounds disagree, kept cheap by a vectorized tree build (31 → 2 ms, 300-seed byte-identical equivalence test) and per-candidate caching in the Pareto beam (28 → 5 ms, identical results)); `dp_rank_forward` falls back to the memory-minimal layout before refusing; output-head chunking and memory margins are *internal calibrated constants*, not planner decisions; `dp_rank_forward` plans once and raises when the unsplit plan cannot be admitted. Best-effort internal splitting and planner-driven head/memory decisions are scheduled follow-ups.

**Breaking change:** `shared_prefix_max_depth`, `head_chunk_tokens`, `memory_safety_factor`, and `memory_reserve_fraction` are removed with no deprecation aliases — passing them is an immediate `TypeError`. Migration: delete the arguments. `TrainerRankMemoryError` now carries `predicted_peak_bytes`, `usable_limit_bytes`, and `suggestion`. TP>1 raises `TrainerRankRuntimeSupportError` (see Known limitations).

## Design

- **Planner core** (`_prefix_tree_planner.py`, `_planner_cost.py`, `_prefix_tree_performance_search.py`): canonical full radix tree (no depth ceiling), a deterministic mandatory candidate family (no-sharing / depth-1 / full-sharing anchors, all uniform depths, effective-span breakpoints, savings plateaus), a calibrated integer cost model, and a bounded Pareto-beam refinement search that is oracle-exact on ≤14-decision trees. Adopted from the validated research implementation (frozen 2026-08-31) with research-only surfaces removed.
- **Integration** (`_impl.py`): per checkpoint-group, selection is cached by content identity and materialized into the existing `PrefixTreePack` execution path — no Megatron changes were needed for arbitrary-depth layouts. Head chunking and memory margins are internal calibrated policy.
- **Distributed identity without a leader protocol**: selection is a pure function of (content, topology, coefficient version) and never reads rank-local memory facts, so ranks in a replica can't diverge; memory/width decisions consume already-all-reduced facts. Zero new collectives.
- **Overlapped planning**: `forward_micro_batches` speculatively pre-plans the predicted next wave (this DP rank's strided slice) on a background thread during the caller's GPU time, from immutable CPU token snapshots taken on the calling thread. Selection is a pure memoized function, so speculation can never change a plan and cannot be poisoned by caller-side tensor mutation; CUDA inputs skip it. Synchronous submission cost is charged to `planning_ms`; hidden worker time is `speculative_planning_ms`. Honest measurement (2-layer, 2-wave, fresh tokens): ~1–2 ms/step of critical-path planning reclaimed and wall time neutral — sharing-aware width pricing already plans the accepted width, so this is insurance for planning-heavy regimes rather than a throughput feature.
- **Telemetry & test hooks**: `last_forward_telemetry()` (`selected_max_depth`, `planning_ms`); env-gated `ART_TRAINER_RANK_TEST_ANCHOR` layout forcing for paired measurement (inert in production, never a public argument).

Full design brief: `dev/trainer_rank_planner_design.md`. Acceptance criteria and provenance: `dev/trainer_rank_landing_acceptance_README.md`.

## Evidence (all acceptance gates pass)

Written test-first (33 acceptance tests + GPU validation scripts, fail-verified against `main` before implementation). Gates derive from the research thread's sealed final-acceptance campaign.

| Cell | Gate | Sealed research | This PR (measured) |
| --- | --- | --- | --- |
| GRPO GDN CP4 (4×H200, 2-layer screen) | paired median gain ≥20% | +47.2% | **+49.4%** |
| 〃 | peak reduction ≥30% | −55.8% | **−53.7%** (4.17 vs 9.0 GiB) |
| 〃 | median selected depth ≥2 | 3 | **3** |
| 〃 | planning p50 ≤150 ms | 82 ms | **37 ms** (all cache-miss) |
| CP1 full-height, real branchy stream | regression ≤2% | tie | **−0.2%** (tie) |
| 〃 | planning fraction ≤10% | 1.5–4.2% | **1.4%** |
| CPU census (44 real groups) | 0 refusals | 0 | **0** |
| GPU anchor parity (H200) | outputs match across no-sharing/depth-1/full-sharing/automatic | — | worst 8e-4% (bf16 noise); head/slot backward exact |

Regression: trainer-rank unit shard 177 passed (knob tests converted to knob-rejection tests, each change justified below); broad unit sweep at baseline; `prek run --all-files` green. Acceptance tests are wired into the CI Megatron shard.

## Test changes

- `test_trainer_rank_accepts_shared_prefix_depth` → `test_trainer_rank_rejects_removed_planner_knobs` (the knob is gone by design).
- `test_trainer_rank_weird_shapes.py`: constructor knob args removed from setup (assertions unchanged — they test invariants, not depth policy); OOM-context test asserts the new structured error fields.
- `dev/trainer_rank_check.py` and `scripts/ci/trainer-rank-gpu-tests.sh`: forced depths/chunks → forced anchors via the test hook.

## Known limitations (documented, follow-ups)

1. **No internal splitting in `dp_rank_forward`** (same as `main`): it plans once and raises `TrainerRankMemoryError` if the unsplit plan cannot be admitted. Best-effort splitting is a follow-up PR with its own memory-pressure gate.
2. **Head chunking / memory margins are constants** (`_HEAD_CHUNK_TOKENS`, `_MEMORY_SAFETY_FACTOR`, `_MEMORY_RESERVE_FRACTION`), revisited as planner decisions later.
3. **TP>1 refuses** with `TrainerRankRuntimeSupportError`: the planner's admission model is not TP-calibrated. Follow-up: widen and re-run topology acceptance.
4. **Cost model undervalues full sharing on some GRPO cells** (sealed research: full-sharing arm faster than automatic on the win cell). Constants are versioned (`COEFFICIENT_VERSION`) for recalibration against production telemetry.
5. The real-workload census corpus is git-ignored (customer-derived tokens); the census phase requires it locally.

### DIFF
--- a/.github/workflows/prek.yml
+++ b/.github/workflows/prek.yml
@@ -225,6 +225,7 @@ jobs:
             tests/unit/test_prefix_tree_packing.py \
             tests/unit/test_trainer_rank_validation.py \
             tests/unit/test_trainer_rank_weird_shapes.py \
+            tests/acceptance/trainer_rank_planner \
             tests/integration/megatron/gdn_shared_prefix/test_gdn_planner_runtime_model.py \
             tests/integration/megatron/gdn_shared_prefix/test_gdn_cp_layout_distributed.py::test_distributed_gdn_cp_layout_all_to_all_roundtrips \
             tests/integration/megatron/gdn_shared_prefix/test_gdn_cp_layout_distributed.py::test_distributed_gdn_cp_layout_handles_empty_ranks
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ replays/
 /src/art/wandb/__pycache__/
 scratch/
 /progress_log.md
+dev/_trainer_rank_ellavox_qwen35_4b_tokens.json
--- a/dev/trainer_rank_check.py
+++ b/dev/trainer_rank_check.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from collections.abc import Iterable, Sequence
+from collections.abc import Iterable, Iterator, Sequence
+from contextlib import contextmanager
 from dataclasses import dataclass
 import json
 import os
@@ -13,6 +14,26 @@ import torch.distributed as dist
 from trainer_rank_diag import all_ranks_checked, rank0_checked
 from trainer_rank_support import load_random_checkpoints
 import typer
+
+ANCHOR_HOOKS_ENV = "ART_TRAINER_RANK_TEST_HOOKS"
+ANCHOR_ENV = "ART_TRAINER_RANK_TEST_ANCHOR"
+
+
+@contextmanager
+def _forced_anchor(anchor: str) -> Iterator[None]:
+    """Force one layout anchor via the test-only planner hook."""
+
+    if anchor == "automatic":
+        yield
+        return
+    os.environ[ANCHOR_HOOKS_ENV] = "1"
+    os.environ[ANCHOR_ENV] = anchor
+    try:
+        yield
+    finally:
+        os.environ.pop(ANCHOR_HOOKS_ENV, None)
+        os.environ.pop(ANCHOR_ENV, None)
+
 
 from art.megatron.prefix_tree_packing import prefix_tree_pack
 from art.trainer_rank import (
@@ -44,9 +65,8 @@ def main(
     mode: Literal["correctness", "performance"] = "correctness",
     model: str = "Qwen/Qwen3-0.6B",
     layers: int = 1,
-    depths: str = "0,1,2,3,4",
-    performance_depth: int = 1,
-    chunks: str = "17,512,8192",
+    anchors: str = "no_sharing,depth_one,full_sharing,automatic",
+    performance_anchor: str = "automatic",
     workload: Literal["regular", "austin", "varied", "unequal_slots"] = "regular",
     request: Literal["target", "multi", "topk", "logits", "hidden", "mixed"] = "target",
     families: int = 8,
@@ -88,15 +108,14 @@ def main(
             payload = _correctness(
                 runtime,
                 base_model=model,
-                depths=_ints(depths),
-                chunks=_ints(chunks),
+                anchors=tuple(anchors.split(",")),
                 slots=slots,
             )
         else:
             payload = _performance(
                 runtime,
                 base_model=model,
-                depth=performance_depth,
+                anchor=performance_anchor,
                 workload=workload,
                 request=request,
                 families=families,
@@ -125,11 +144,10 @@ def _correctness(
     runtime: Any,
     *,
     base_model: str,
-    depths: tuple[int, ...],
-    chunks: tuple[int, ...],
+    anchors: tuple[str, ...],
     slots: int,
 ) -> dict[str, object]:
-    assert depths and chunks, "depths and chunks must not be empty"
+    assert anchors, "anchors must not be empty"
     rank = TrainerRank(runtime)
     slot_names = load_random_checkpoints(
         runtime,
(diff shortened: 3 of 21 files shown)


bradhilton:
## Response to review (research-thread review of c133151)

All four technical findings were legitimate and are fixed in [5c3dee4](https://github.com/OpenPipe/ART/pull/826/commits/5c3dee465) and [d56c9d3](https://github.com/OpenPipe/ART/pull/826/commits/d56c9d343). Scope is corrected as discussed: **this PR is automatic arbitrary-depth prefix layout selection + knob removal**, reusing main's width selector (now sharing-aware); head chunking and memory margins are internal constants; `dp_rank_forward` does not split internally (follow-up PR).

### 1. Speculative planning cache poisoning — fixed
The worker now plans from immutable CPU clones taken on the calling thread — the same bytes that produced the cache key — so caller-side tensor mutation after the yield cannot create a mismatched entry. CUDA inputs skip speculation entirely (worker never touches the device). Regression test reproduces the reported scenario (mutate after first yield → cached entry still matches the original content).

### 2. DP>1 speculation warmed the wrong slice — fixed
The width search's geometry (`_wave_geometry`, `_normalize_wave_width`, `_local_wave_indices`) is now shared, so speculation predicts exactly what the next wave will plan: the largest width so far over this rank's strided slice. Test at DP2 asserts the local key is warm and the global key is not.

### 3. No-sharing estimate rejected widths sharing would fit — fixed
The cheap no-sharing count now only *accepts* widths; when it would reject, the planner's actual layouts are priced (memoized) and the memory check redone. Test reproduces the reported case (two rows sharing 1,000 of 1,001 tokens, budget 1,200 → one wave of 2). On the GPU benchmark the same budget admits **2 waves/step where the bound forced 4**.

Exact pricing is kept cheap by two behavior-preserving optimizations: a vectorized canonical-tree build (31 → 2 ms on the sealed win-cell shape; new 300-seed equivalence test proves byte-identical output vs. the verbatim scalar algorithm) and per-candidate caching of Pareto-beam orderings (28 → 5 ms per search; 210 baseline search fingerprints unchanged). Widths at or above a width whose exact pricing failed skip exact evaluation. I checked whether the refinement budget could be lowered instead — it changes 7/65 real selections, so no.

### 4. `planning_ms` excluded synchronous submission — fixed
Submission cost (grouping, snapshots, hashing) is charged to `planning_ms`; hidden worker time is reported separately as `speculative_planning_ms`; speculation groups without the checkpoint-slot ensure-collective (confirmed it was an unconditional gather). The earlier "10.2 → 4.5 ms" claim is withdrawn: with honest accounting speculation reclaims ~1–2 ms/step and is wall-neutral on the launch-bound benchmark, since exact pricing already plans the accepted width. Kept as insurance for planning-heavy regimes; it is one isolated commit if you'd rather drop it.

### 5. Scope — corrected
PR body, design brief (`dev/trainer_rank_planner_design.md`), acceptance README, and the policy-scan test header now state precisely what is planner-decided vs. internal constant, and list splitting and planner-driven head/memory decisions as follow-ups. Stale "EXPECTED TO FAIL" headers removed.

### Benchmark (2-layer, fresh tokens, forced multi-wave)
| | planning/step | step wall | waves |
| --- | --- | --- | --- |
| no-sharing bound (before fix 3) | 10 ms | 153.5 ms | 4 |
| fix 3 as first landed | 67.7 ms | 185.8 ms | 2 |
| now | 42.3 ms | 160.7 ms | 2 |

Verification: 484 CPU tests (33 acceptance + 302 tree-equivalence + 149 trainer-rank shard), `prek --all-files`, GPU anchor parity identical, sealed GPU gates unaffected (`dp_rank_forward` path).

bradhilton:
## Response to re-review: non-monotone width feasibility — fixed in [65d0e10](https://github.com/OpenPipe/ART/pull/826/commits/65d0e1006)

Confirmed the finding (reproduced exactly: 41 → 82 → 43 packed tokens at widths 1/2/3 under the production cost model). Rather than exploring past intermediate failures heuristically, the fix makes the feasibility predicate monotone by construction:

- **A width is feasible iff its memory-minimal (full-sharing) layout fits.** Full sharing minimizes packed tokens, and its packed count is monotone in width, so the existing exponential/binary search structure (and `exact_failed_width`) is valid again.
- **Admission executes the cost-optimal layout when it fits and the memory-minimal layout otherwise.** The chosen mode is recorded per width and threaded into materialization, so the executed plan is always the one that was priced (no optimistic-check gap).
- **`dp_rank_forward` applies the same fallback before refusing** — a modest best-effort improvement in the direction of the (still follow-up) splitting work.
- Both bounds are cheap O(tokens) walks of the packing primitive (no-sharing and unlimited-depth sharing); planner pricing runs only inside the band where they disagree. The canonical tree is cached by content so both modes of a group share one construction.

**Regression tests added** (`tests/unit/test_trainer_rank_weird_shapes.py`): the fail-at-2/fit-at-3 case (asserts the 82/43 premise under the production model, then one wave of 3 under a 60-token budget), and the `dp_rank_forward` fallback (executes the 42-token full-sharing plan where the 82-token cost-optimal plan would not fit). Three pre-existing width-search tests used fixture rows with a shared 3-token prefix whose budgets the memory-minimal fallback now legitimately exceeds; their rows are changed to have no shareable prefix so packed-token budgets keep their original meaning (assertions unchanged).

**Cost:** the predicate pays for itself — per-step planning on the benchmark dropped from 42.3 ms to **6.1 ms**, and the step is now **129.3 ms vs 153.5 ms** for the original no-sharing-bound 4-wave version (2 waves, sharing-aware).

Verification: 486 CPU tests (acceptance, tree-equivalence, trainer-rank shard), `prek --all-files`, GPU anchor parity identical, adaptive-wave smoke unchanged.

bradhilton:
## Response to review: stale profile-trust inputs — fixed in the latest commit

Confirmed: after exact pricing selected a layout, the trust check still read the initial no-sharing bound's `packed_tokens`/`signature` (a leftover from refactoring the pricing into a closure that returned only the memory check). Fixed by carrying the selected pricing — check, packed tokens, output bytes, signature — out of the pricing helper and using it for both trust and the profiled flag. Exact pricing now also runs when the cheap bound fits memory but fails trust against an existing profile, and a candidate is accepted only when it both fits and is trusted (or unprofiled), so the stale bound can no longer reject a width on either axis.

**Regression test** (`test_profiled_steady_state_keeps_the_wide_shared_wave`): GRPO-16 (sixteen rows sharing a 1,000-token prompt; selected plan 1,016 packed vs a 16,016 no-sharing bound), existing profile of 1,016 tokens, 1,100-token budget → one wave of 16, not cold-start. Before the fix this regressed to width 8.

Verification: 487 CPU tests, `prek --all-files`, GPU anchor parity identical, adaptive-wave smoke unchanged; benchmark unchanged at ~6 ms/step planning.

PR #829, opened 2026-09-01
Preserve exact histories across nonterminal length stops

bradhilton:
## Summary

- admit exact projected chat histories across nonterminal length-limited responses only when the renderer proves a unique assistant stop boundary
- verify the complete renderer-owned suffix through the next sampled generation boundary against the next authoritative prompt, preventing cross-role stop-token collisions
- keep terminal, ambiguous, context-changed, explicitly overridden, or otherwise unproven histories on the existing renderer fallback
- rebase onto current `main`, including #830's length-stop provenance semantics

## Provenance semantics

The fully retained raw sampled output remains exact, sampled, and assistant-owned. Preceding non-stop renderer-tail tokens are exact, assistant-owned, and unsampled. The unique template-added terminal boundary token is exactly `EXACT | STOP`, without `ASSISTANT`, matching #830.

## Verification

- `216 passed`: full `tests/unit/trajectories/test_tokenize.py` with the Torch-enabled ART environment
- `211 passed, 5 deselected`: the same focused tokenizer file in the lightweight environment
- `408 passed`: current post-rebase `tests/unit/trajectories` suite
- `1026 passed, 33 skipped`: broad feasible unit suite before the final rebase (optional Megatron/Transformer Engine/Unsloth runtime files excluded locally)
- Ruff format/check, Ty, prek, and `git diff --check` pass
- the post-#830 reviewer blocker is resolved by assigning the terminal renderer-tail token exactly `EXACT | STOP`; all three stale expectations are updated
- independent blocker review: otherwise clean after fixing an omitted-tail/following-user-token collision
- replayed all three captured Qwen3.5-4B histories as the exact final authoritative prompt plus final output:
  - 12,145 tokens / 4,566 sampled / 3 stops
  - 14,678 tokens / 7,027 sampled / 5 stops
  - 30,601 tokens / 22,594 sampled / 9 stops

## Regression coverage

- public raw-token drift across a proven nonterminal length boundary
- omitted assistant boundary colliding with an identical following user special token
- multi-token renderer tails, ambiguous stops, mismatched tails, terminal length fallback, and explicit template override fallback

### DIFF
--- a/src/art/trajectories/_tokenize.py
+++ b/src/art/trajectories/_tokenize.py
@@ -3886,6 +3886,21 @@ def _rendered_length_stop_boundary(
     )
 
 
+def _next_assistant_span_start(
+    assistant_mask: Sequence[bool], *, after: int
+) -> int | None:
+    if not 0 <= after < len(assistant_mask):
+        return None
+    return next(
+        (
+            index
+            for index in range(after + 1, len(assistant_mask))
+            if assistant_mask[index] and not assistant_mask[index - 1]
+        ),
+        None,
+    )
+
+
 def _tokenize_exact_projected_chat_history(
     history: ChatCompletionsHistory,
     *,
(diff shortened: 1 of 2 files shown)


bradhilton:
Addressed the review blocker by making a proven nonterminal length-stop terminator exactly `EXACT | STOP` (never `ASSISTANT`) and updating the three stale expectations.

A checkpoint-120 live replay then exposed an additional combined #829/#830 case: all trajectories ended in a terminal length stop, so the exact path fell back wholesale and failed to locate an intervening reasoning+content response. Commit 0cba3b752 keeps earlier history from the authoritative final prompt and appends only the renderer-certified terminal tail with the same synthetic provenance as ordinary rendering. It also finds a missing next-message bound from the next assistant-mask transition; the subsequent exact-prompt check remains authoritative.

Validation so far: 217/217 tokenizer tests, Ruff, format, ty, diff check, and offline tensorization of both captured live groups (14 trajectories; 419,831 total tokens) pass. I am holding merge until CI and a new live checkpoint-120 validation replay pass.

bradhilton:
Final live validation passed on `0cba3b752dd72befa283c4f994210b9a55c6030d` from checkpoint 120. The replay tensorized both judged groups, completed a successful trainer update (`ESS=16`, zero rollout errors/history splits), published and saved checkpoint 20, completed all 80 validation trajectories, ran checkpoint retention, and exited cleanly. W&B run: https://wandb.ai/wandb/tau2-bench-overfit-val-main/runs/046-tau2-bench-overfit-val-new-bmQwen-Qwen3.5-4B-mt12-jmgpt-5.6-luna-jrehigh-ls0.25-h1e1fef08d49c852f

All PR checks are green; proceeding with the authorized merge.

PR #832, opened 2026-09-02
Admit tensor-parallel runtimes in TrainerRank (TP>1 support)

bradhilton:
## Summary

Lifts the TP>1 constructor refusal that #826 added as a calibration caution. TrainerRank accepted tensor parallelism before the automatic planner landed, and all of that machinery is unchanged and pre-dates the planner: the vocab-parallel output head (log-Z, target logprobs, top-k and full logits reduced/gathered across the sharded vocabulary), the sequence-parallel hidden gather and output-layer SP toggle, TP padding of packed batches, sharded vs replicated LoRA gradient reduction, memory checks all-reduced within the TP×CP group, and memory-profile signatures keyed by (dp, tp, cp, pp) so a TP topology calibrates itself online. PP>1 (and multi-chunk runtimes) still refuse.

One production change beyond the lift, from the research thread's review: execution pads every checkpoint/no-grad group independently to a TP multiple, so a plan's `packed_tokens` (and with it the cheap bounds, the split lower bound and the memory profile) is now the *physical* count with each group rounded up; the unpadded aggregate could be short by `groups × (TP−1)` tokens, not merely `< TP`.

What had never happened before this PR: the planner executing end to end under TP through the public API. The research thread's only TP2 attempt stopped at a research-only preflight; DP2×TP2 had never been run at all. Both are now gated, and both pass.

## Gates (test-first; every gate failed on the refusing tree at `TrainerRank(runtime)`)

- `tests/unit/test_trainer_rank_topology.py` (CPU): TP>1 runtimes construct; PP>1 and multi-chunk runtimes still refuse.
- `dev/trainer_rank_landing_acceptance.py --phase tp2-public --tp 2` (2× H200, Qwen3.5-4B full model, DP1×TP2×CP1, public `dp_rank_forward`, active LoRA slot) plus the identical cell at `--tp 1` as the control: both TP peers plan the same physical layout on every call; the automatic planner shares more deeply than depth-one on the hierarchical GRPO shape; odd packed lengths exercise sequence-parallel padding with outputs at the final real token; losses finite; measured rows compile-free and plan-cache-stable (3 warmups per arm, 10 alternating measured pairs); paired timing reported, not gated.
- `--phase tp-compare` (CPU) gates numerics on the two dumps. bf16 rounding differs whenever the reduction order changes, and both a different packing and a different TP degree change it; over 36 layers that noise is ~1.3% of the mean logprob magnitude (0.19–0.20 nats per target token) on this cell, so absolute tolerances borrowed from same-packing comparisons cannot separate a TP defect from noise (the first TP2 run failed exactly that way). The reference is the control's own cross-layout divergence measured in the same run: same-layout TP2-vs-TP1 divergence (mean and max) and cross-layout divergence at TP2 (outputs and LoRA gradients) must stay within 1.5× of it, losses must agree, and differences must be unstructured (no request outlier, final tokens like the body, no bias).
- `--phase dp2-tp2-waves` (4× H200, DP2×TP2, public `forward_micro_batches`, active LoRA slot): at least two waves under the test-only memory cap, DP replicas with different payloads, identical wave shapes within each TP pair, every global input returned exactly once in original order, forward and backward per wave, automatic vs depth-one parity, and an empty-DP-slot arm; completing is the no-hang gate (world-scope planning collectives composed with pair-scope TP execution collectives).
- CI: `dev/trainer_rank_check.py` at TP=2 (16 request combinations, two LoRA slots, no_sharing vs full_sharing) next to the existing CP=2 run; the GDN TP2 kernel-parity test now also runs with LoRA (its TP2 arm was base-model only).

## Evidence (Kubernetes H200s and one local H200; final phase code)

| Cell | Result |
| --- | --- |
| DP1×TP2 public cell, layouts | automatic depth 3 with a **9,199-token** group vs depth-one 37,871, for 53,216 logical tokens (both odd, so SP padding is exercised; physical counts 9,200 / 37,872 after the per-group padding fix); TP peers identical on every call; same-layout run-to-run: outputs bit-identical, LoRA grads 1.3% rel-L2 (bf16 backward is nondeterministic) |
| DP1×TP2 timing (10 compile-free, cache-stable pairs) | automatic **720 ms** vs depth-one 1,921 ms per forward+backward, **62.5% paired median gain** (TP1 control: 809 vs 2,878 ms, 71.9%) |
| tp-compare, same layout TP2 vs TP1 (same unpadded group lengths, matching source and workload fingerprints) | automatic 1.39% mean (ratio **1.06** to the 1.31% reference), max-abs 1.66 (reference 2.44); depth-one 1.38% (ratio 1.05), max-abs 1.61; losses within 0.009% / 0.061%; flat per-request profile (0.18–0.23 nats), tail = body, signed mean ≈ 0 |
| tp-compare, cross layout at TP2 vs TP1 | outputs 1.37% vs 1.31% (ratio **1.05**); LoRA gradients rel-L2 0.134 vs 0.127 (ratio **1.06**) |
| DP2×TP2 waves | 2 waves (4 then 2 items; 10,222 packed for 53,216 logical per rank), DP ranks held items {0,2,4} vs {1,3,5}, TP pairs identical, every input once in order, empty-slot arm: one DP replica ran the single item, the other ran the wave with zero outputs; automatic vs depth-one outputs identical (single-group items coincide), grads 1.17% rel-L2 (the run-to-run floor) |
| CI 2× H200 | check script at TP=2: 0.0 divergence on outputs, head backward and slot backward; GPU shard 56 passed with the GDN TP2 LoRA arm |

All GPU rows are from the final source (the k8s cells were rerun after the review fixes); both clusters were torn down after collecting the logs and dumps.

## Known, accepted limitations

- The cost model carries CP terms but no TP terms. Layout ranking survived here (the depth-3 layout wins 62% at TP2), but constants are uncalibrated; TP terms belong to the cost-model recalibration follow-up and should be fitted from fresh TP2 telemetry, not by dividing the estimate by TP (some tensors are replicated or gathered before the head, so a blanket division could turn conservative refusals into unsafe admissions).
- The cold static estimate ignores sharding (conservative: a cold oversized call may refuse until profiled; no cold refusal occurred in these cells).
- The all-shards-`-inf` output-head case is a documented non-goal: unreachable for supported models without a vocabulary-wide mask.

Design brief: `dev/trainer_rank_planner_design.md` ("Tensor parallelism").


bradhilton:
Gate results on the final source (both k8s clusters torn down afterwards; artifacts local):

- **DP2×TP2 waves**: passed on the first run and on the rerun: 2 waves (4 then 2 items), TP pairs identical, DP replicas held disjoint items, every input returned once in order, forward+backward per wave through an active LoRA slot, empty-DP-slot arm correct.
- **DP1×TP2 public cell**: structural gates pass (identical layouts on both TP peers, automatic depth 3 with 9,199 packed tokens vs depth-one 37,871 for 53,216 logical, odd lengths so SP padding is exercised, compile-free and plan-cache-stable measured rows). Timing: automatic 722 ms vs depth-one 1,885 ms per forward+backward, 61.9% paired median gain (TP1 control: 71.9%).
- **Numerics**: the first TP2 run failed the absolute parity tolerance I had borrowed from the same-packing split gate (1.37% mean divergence automatic vs depth-one). The TP1 control of the identical cell shows the same divergence (1.31%): bf16 rounding differs whenever the reduction order changes, and a different packing and a different TP degree both change it, so over 36 layers the floor is ~0.19–0.20 nats per target token either way. Same-layout TP2-vs-TP1 divergence is 1.39% (ratio 1.06 to that reference) with a flat per-request profile, final tokens like the body, no bias, losses within 0.06%, and a smaller max than the cross-layout reference. The new `--phase tp-compare` gates relative to the control's cross-layout divergence (≤1.5×) plus structured-error checks, and passes. bf16 backward is nondeterministic: same-layout LoRA gradients differ ~1.2–1.4% rel-L2 run to run, which is why the gradient comparison is also relative.
- **CI (2× H200)**: `dev/trainer_rank_check.py` at TP=2 shows 0.0 divergence on outputs, head backward and slot backward; GDN TP2 LoRA arm passes.

Ready for the research thread's review.

bradhilton:
Both review items are fixed, and all GPU evidence was regenerated on the final source.

**P1 (per-group TP padding in admission)** — `_FlatForwardPlan.packed_tokens`, both cheap estimate paths (width probing) and the split lower bound now use the physical count: each group's packed length rounded up to the topology's TP (`_physical_tokens`), matching what `_pad_packed_batch` does per group at execution. The memory profile learns from the same count. Unit test: two odd-length groups (5 and 7 tokens) count 6 + 8 at TP=2 across the plan, the cheap and exact estimates and the split lower bound; the unpadded sum at TP=1. The TP2 cell now reports physical 9,200 / 37,872 for group lengths 9,199 / 37,871.

**P2 (compare gate must exercise TP)** — `tp-compare` requires the trial dump to come from the expected TP (>1) cell and both dumps to carry the same source and workload fingerprints; supplying the TP1 dump as the trial is rejected, and stale dumps are rejected. Two details worth knowing: the source fingerprint is a content hash of `src/art/trainer_rank/*.py` rather than a git commit (SkyPilot syncs the workdir without `.git`, so a commit-based fingerprint read "unknown" on the cluster); the driver's own hash is recorded for information only, since its gate logic does not change numerics. "Same layout" is decided on the unpadded per-group lengths, because `packed_tokens` is now physical and differs by the padding between TP1 and TP2.

**Final-source evidence**: DP2×TP2 waves passed (2 waves, disjoint DP payloads, identical TP-pair shapes, every input once in order, empty-slot arm). TP2 public cell passed (TP peers identical, depth 3, SP padding exercised, compile-free cache-stable rows; automatic 720 ms vs depth-one 1,921 ms, 62.5% paired gain). tp-compare passed: same-layout TP2-vs-TP1 ratios 1.06 / 1.05 to the TP1 cross-layout reference, cross-layout ratios 1.05 (outputs) / 1.06 (LoRA grads), losses within 0.06%, unstructured differences. CI check script at TP=2: 0.0 divergence.

Residual gap noted as non-blocking, as you said: TP×CP and TP4 remain untested; evidence covers TP2×CP1 and DP2×TP2.

PR #864, opened 2026-09-08
Support filtered trajectory copies and safe empty-rank forwarding

bradhilton:
Uneven data-parallel batches can give a rank an empty microbatch, making an otherwise ordinary loss/backward loop fail. `TrainerRank.forward_micro_batches(..., yield_empty=False)` now skips locally empty yields by default while every rank still participates in internal planning and forward operations. `yield_empty=True` retains the previous behavior. Public collective operations reject calls from waves that skip ranks; fully participating waves allow them. Guards handle nested and retained iterators, with documented thread scope and cleanup requirements.

Add `include_models` and `exclude_models` to `Trajectory.model_copy()` and `TrajectoryExchanges.model_copy()` so callers can copy the dynamics exchanges directly. Selectors prefer exact captured model names, then case-sensitive shell patterns. Copies get fresh exchange lists; `deep` controls retained object copying, and explicit `update` values preserve Pydantic's existing semantics. Legacy message history remains unchanged by model filtering.

Checkpoint prefetch now waits for backing-source preparation even when a snapshot is already resident. This prevents callers from cleaning the materialization pin while preparation is pending and keeps the source usable when a later save evicts that snapshot. The resident-snapshot, delayed-preparation, cancellation, and eviction regression passes; all 17 targeted prefetch tests pass.

Snapshot names now retain their forward-only identity across resident eviction and registered reload. Only names created through snapshot helpers are marked; ordinary logical or exact-name training loads remain mutable. This prevents Caladan’s default grouped save from including an old validation snapshot as a second member of the same collection. The correction adds six net production lines; 32 targeted tests pass, including actual LoRA/custom-state reloads for both snapshot origins and either prefetch-registration order.

Final CI passed: 714 lightweight tests and 1,069 unit tests (21 skipped), plus 61 GPU tests (one skipped) and CP2/TP2 public-API validation on two H200s. Integration fixtures initialize the same private guard/snapshot state and locks as production constructors. Ruff, formatting, type checks, and lock validation pass; both independent reviewers cleared the final head.

Caladan’s [targeted H200 qualification](https://github.com/OpenPipe/caladan/pull/239#issuecomment-5591939032) additionally confirms both snapshot origins reload frozen, ordinary exact-name training loads remain mutable, and default grouped selection excludes saved snapshots. All owned validation pods and services were cleaned up. The squash-merge tree is identical to the reviewed, CI-tested head.

PR #859, opened 2026-09-07
CP planner: price host work per remote stage, real fetch bandwidth and per-token compute (#854)

bradhilton:
Fixes #854.

## What the planner got wrong

The context-parallel assignment planner (`_search_generic_chunk_assignment`) chooses the remote wave count and the chunk owners from its own per-layer cost model. Two of its constants and one missing term made it systematically choose slower plans:

1. **No host cost per remote stage.** Every remote wave adds a fixed slab of host work per layer (its own attention call, merge and collectives in forward and backward). Measured as the extra time of one more wave at equal max rank load on H200 bf16 (43 pairs per class): +2.3–2.6 ms/layer on Qwen3-1.7B, 4B, 8B (CP4) and 30B-A3B; +0.2–0.6 ms where layers are long enough (≥35 ms) to hide it. The planner predicted −0.2 ms and preferred the extra wave in 33/43 pairs.
2. **Fetch priced at ~14 GB/s.** `planner_fetch_token_ms = 0.000287` prices a 12k-token KV fetch at 3.8 ms, about ten times NVLink. The planner hid that phantom latency behind extra waves and, at CP2, traded token balance for less remote fetch.
3. **Attention only.** The model had no term for the rest of the layer's compute per owned token (projections, MLP or routed experts), so on causal rows it balanced attention pairs and handed the early rank several times the tokens of the last (e.g. 14,336 vs 4,635 tokens at TP2×CP2 on a measured Ellavox group).

## Change

- `planner_remote_stage_host_ms` (new, 1.2 ms per direction) added to remote stage costs.
- `planner_fetch_token_ms` / `planner_reduce_token_ms`: 0.000287 → 0.00003 (30 ns/token, a conservative all-to-all figure for 4 KB tokens on NVLink).
- `planner_owned_token_ms` (new): per-owned-token compute from the provider geometry, `estimate_owned_token_ms` = 3 × (8h² + 6h·ffn_eff) / TP / 400 TFLOP/s; set by `_context_parallel_config_for_provider`. Agrees within ~30% with the per-token coefficients fitted from measured layer times on four classes (0.7 / 2.8 / 4.6 / 1.3 µs on 1.7B / 8B / 14B / 30B-A3B). The evaluator adds owned tokens × cost per rank (⅓ forward, ⅔ backward).

[...]

PR #884, opened 2026-09-10
GDN planner: price the dense projection work per GDN-owned token; refit the chain overheads and gate

bradhilton:
Follow-up to #859 (the GDN-planner item of its follow-up list).

## What the GDN planner got wrong

After the attention-to-GDN all-to-all the whole GDN layer — input projection, convolution, recurrence, output projection — runs on the GDN layout, so a rank pays the projections for every GDN token it owns, chained or not. The GDN planner's runtime model (`GdnPlannerConfig`) priced a rank's work by a recurrent rate alone (1,500 tokens/ms), fitted once on the 35B reference shape. On the paired A/B of #859 this showed as a 23% slowdown of one Qwen3.5-4B layout: the planner stopped chaining a 12.6k-token sequence across the ranks because it predicted a 4 ms saving against a 4 ms gate, while the measured saving was about 15 ms per layer.

## Measurement

`--gdn-ab` times every layout of a cell under the production GDN planner and under two forced decisions that bracket its choice — never chain, chain every legal segment — in alternating rounds on the same node with the attention planner unchanged, so the paired difference is the GDN decision's own cost. Campaigns on Qwen3.5-4B (CP2, CP4), Qwen3.5-27B (CP2, CP4) and Qwen3.5-35B-A3B (CP2 EP1/EP2, CP4 EP2/EP4): 982 layouts. Chaining everything is the wrong move on most layouts (short segments; the chain overhead dominates), but where it wins the model missed it by 3–70×: one 12.5k-token sequence at CP4 measures 16.7 ms per GDN layer faster chained on 4B (predicted 3.8) and 48 ms on 27B.

## Change

- `runtime_dense_tokens_per_ms`: the projection work per GDN-owned token, from the model shape's FLOPs at the throughput the three campaigns fit jointly (156 TFLOP/s; 617 tokens/ms on the 4B shape, 224 on 27B, 772 on the 35B reference). It enters the per-rank critical path of the runtime prediction and the chain scoring; the owner search balances recurrent work only (pricing the dense work there moved local segments toward token balance at the cost of more layout exchange, measured as noise with a +14% outlier).
- Chain overheads refit with the recurrent rates and exchange costs held at their shipped values: 2.2 ms per bucket launch (was 0.2), suffix scan 2.8 ms per bucket plus 3.5 segments/ms at the reference shape (was 2.0 and 15); the per-byte summary-exchange cost is not supported by the data and retires into the per-segment scan.
- Chain gate 4.0 → 2.0 ms: every chain that measured slower on 4B was predicted to save under 2 ms per layer; 2 ms keeps 97% of the measured gain on 4B and 99% on 27B.
- Model error on the paired deltas 10.1 → 5.2 ms per layer (rms); the regret of choosing between the two arms by the model 1,082 → 297 ms per layer summed over 982 layouts.

## Validation against main's GDN planner (paired, same node, attention planner unchanged)

| class | layouts | chain decision changed | faster > 2% | slower > 2% | unchanged layouts | production selection |
| --- | --- | --- | --- | --- | --- | --- |
| Qwen3.5-4B | 274 | 72, median −15.2% | 69 | 1 (+3.0%) | 174, ±0.5% | −6.9% at CP4, 0.0% at CP2 |
| Qwen3.5-27B | 274 | 95, median −17.8% | 90 | 1 | 151, ±0.3% | 0.0% (production picks were already chained at CP4) |
| Qwen3.5-35B-A3B | 574 | 119, median −3.7% | 70 | 11 (worst +6.5%) | 371 (+30 synthetic CP4 EP1), ±1% | +0.5% to −0.3% per shape |

Best layout per real-data group: −15% to −43% at CP4 and −21% to −34% at CP2 on 4B; up to −30% per layout on 27B; within ±1% on 35B. **Known residual:** the model over-predicts chain savings on 35B by about 2× (11 of 119 changed layouts slower by more than 2%).

## Re-certification

| table | cells | pairwise | p95 regret | max regret | before |
| --- | --- | --- | --- | --- | --- |
| dense-h2560 (Qwen3.5-4B, Qwen3-4B) | 58 | 99.0% | 1.2% | 1.9% | 98.7% / 1.9% / 2.8% |
| gdn-moe-h2048 (Qwen3.5-35B-A3B) | 80 | 97.8% | 2.2% | 4.2% | 96.6% / 2.7% / 4.2% |
| dense-gdn-h5120 (Qwen3.5-27B) | **not refit** | | | | 98.6% / 3.9% / 4.6% kept |

Every admitted CP > 1 cell of the three GDN tables was re-measured by the validation campaign (its current-planner rows); the fitter now re-certifies from new evidence plus the previous certificate's aggregates for the cells a change does not touch, named explicitly (`--from-certificate` with evidence, `--carry-shapes` for the CP1 and TP2 × CP1 shapes, `--carry-cells` for the attention-only Qwen3-4B geometry on the dense table); any other certificate cell missing from the evidence refuses the re-certification, and the lists are recorded in the certificate. **The 27B refit fails its held-out gate on one cell** (Ellavox g3 at CP4, 11.3%): the deep `uniform_depth_3` layout's long GDN segments now chain and it becomes 10% faster than the shallow layouts the ten terms prefer; neither a CP-split of the GDN level term nor a fit on the CP > 1 cells alone recovers it. The shipped 27B table is kept: under the new planner its picks are within 2.7% of the new best on the other 13 CP4 cells and its g3 pick is 22% faster than before (regret against a new best, not a regression), and withholding the shape would hand the group to version 1, which loses up to 33% there. A second stage pricing shortlisted layouts with the planners' own calibrated models fixes g3 (4.0%) but mis-ranks two synthetic CP4 cells by 7–8% in its pure form; a fitted GDN-aware re-ranker is the follow-up.

## Footprint in `art.megatron`

One file, `art/megatron/gdn/gdn_prefix_tree.py`, about 55 non-comment lines: a new `GdnPlannerConfig` field (`runtime_dense_tokens_per_ms`) derived in `from_model_shape`, one addend in the per-rank critical path of the runtime prediction (threaded into the chain scoring), and refit default constants. It is planning-time CPU code that only changes *which* segments the GDN planner chains at CP > 1; the owner assignment, the plan materialization, the exchange plans, the kernels and every CP1 path are untouched (`_select_chain_segment_keys` returns before any scoring at CP1). Chained execution is a pre-existing production path (main already chains the 27B and 35B production picks at CP4) and is CI-gated by `test_gdn_cp_tree_chain_matches_cp1_oracle` on the 2×H200 job. Across the 1,092 validated layouts the training loss is identical between the two planners wherever the decision is unchanged and differs by at most 1.0e-3 (4B), 6.7e-3 (27B) and 1.6e-2 (35B, MoE routing) relative where a different plan changes the bf16 reduction order — the same plan-dependence as any CP re-plan. The planner constants cannot be set from `art.trainer_rank`: the GDN planner runs inside the Megatron runtime's planning bundle with a config built from the provider, and the dense term has no equivalent among the existing constants (collapsing it into the recurrent rates fits 8.99 vs 5.2 ms per layer rms and over-penalizes short segments).

## Also in this PR

- Harness: `--gdn-ab`, `--gdn-legacy-ab`; the contract phase admits the `yield_empty` flag of #864.
- Fitter: re-certification from evidence plus certificate aggregates for explicitly named unchanged configurations; missing affected cells refuse the run (manifest fingerprints, completeness and export aware).
- Regression test: an 8k-token sequence at CP4 chains under the dense term and stays on one rank without it.

## Follow-ups

- GDN-aware second stage for the layout selector (27B CP4 g3; the 35B chain tail).
- 35B: chain overheads under expert parallelism (the model's 2× over-prediction there).


bradhilton:
Reply to the review of head `98659ee9` (fix in 19f0e8a8a, docs in 0310f6c2c):

**P2, silent carry.** Agreed; the carry is now explicit. `--from-certificate` with evidence requires the configurations eligible for reuse to be named: `--carry-shapes` (parallel shapes the change does not touch) and/or `--carry-cells` (cell-key substrings, e.g. the attention-only Qwen3-4B geometry, which a GDN planner change cannot reach). Every other certificate cell must be present in the evidence or the run is refused with the missing cells listed; both lists are recorded in the certificate's `fit_arguments`. Your two cases are tests now: an empty evidence file refuses with every affected cell named, and one re-measured cell refuses with the other affected cells named; a configuration not on either list is never carried, and a named cell that *is* re-measured comes from the rows. The dense-h2560 and gdn-moe-h2048 certificates are re-exported through this path (carry: `tp1cp1,tp2cp1` + `|Qwen/Qwen3-4B|`; `tp1cp1,tp2cp1,tp2cp1ep2`); the tables are unchanged.

**Footprint.** The PR body gains a section on the `art.megatron` change (one file, ~55 non-comment lines, planning-time only, CP1 paths untouched, chained execution pre-existing and CI-gated) and on the training-loss agreement between the two planners across the 1,092 validated layouts.

bradhilton:

**Change.** Follow-up to #859: the GDN planner now prices dense projection work per GDN-owned token (`runtime_dense_tokens_per_ms` in `gdn_prefix_tree.py`), refits the chain overheads and lowers the chain gate from 4.0 to 2.0 ms. Paired model error fell 10.1 to 5.2 ms per layer over 982 layouts; dense-h2560 and gdn-moe-h2048 were re-certified, the 27B table kept.

**Review.** The record keeps one author reply to a review of an earlier head (`98659ee`); the reviewer is not named. Its P2 request was resolved before the merged head `390a562`. Chained execution is CI-gated by `test_gdn_cp_tree_chain_matches_cp1_oracle`, and a CP4 regression test was added.

**Findings addressed during review.**
- Silent carry of certificate cells during re-certification made explicit via `--carry-shapes` / `--carry-cells`; missing affected cells refuse the run, with tests.
- PR body gained an `art.megatron` footprint section and training-loss agreement across 1,092 layouts.

**Deferred / follow-ups.**
- GDN-aware second stage for the layout selector (27B CP4 g3; 35B chain tail).
- 35B chain overheads under expert parallelism.

PR #912, opened 2026-09-16
Return full source-order context-parallel trainer outputs

### LINKED ISSUE #911: Context-parallel forward must return full-sequence outputs in caller order (ForwardOutput fields are currently rank-local shards)
## Summary

With context parallelism (the default `TrainerRank` topology for dense handlers when more than one GPU is visible: TP 1, CP = visible GPUs), every per-position field of `ForwardOutput` (`hidden_states`, `target_logprobs`, `top_k`, `logits`) is the rank's **local shard** of the sequence, in rank-plan order, not the caller's sequence. The API contract everywhere else (single GPU, DP, TP with sequence parallel) is that a `ForwardOutput` corresponds to the caller's `ForwardInput` in the caller's order. CP silently breaks that contract; callers that index outputs by position fail (or, worse, would compute on wrong positions if shapes happened to match).

**Requested fix: make the CP forward return full-sequence outputs in source order, internally.** Callers should not have to map positions or know the topology.

## Observed

Caladan probe doing `output.hidden_states[1:][mask]` with a full-sequence mask on Qwen/Qwen3.8-27B, 20,436-token input:

- CP 2: `IndexError: The shape of the mask [20435] at index 0 does not match the shape of the indexed tensor [8659, 5120]`
- CP 4: same, `[6655, 5120]`

Shards are uneven (prefix-tree rank plan), so nothing can be reconstructed caller-side without the plan.

## Where it comes from

`_prepare_context_parallel_forward` dispatches `arange(seq_len)` through the rank plan and records `positions_by_item` (local rows) and `source_positions_by_item` (the source positions this rank owns). `_project_head` returns `hidden_states = _select_positions(hidden_by_row, positions)` and sizes `target_logprobs[index]` from `labels.index_select(0, source_positions)`, i.e. local rows only. `source_positions_by_item` is used internally for labels and then discarded.

## Proposed internal fix

Do for CP what `_gather_sequence_parallel_hidden` already does for TP sequence parallelism: gather each item's rows across the CP group and place them in source order before building `ForwardOutput`.

1. For each item, all-gather `(source_positions, values)` over `ps.get_context_parallel_group()`. Shards are unequal, so pad to the rank plan's `pad_multiple` (or the max shard length) and drop the padding after the gather; ART already tracks the padded layout in the rank plan.
2. Scatter the gathered rows into a full-length tensor by `source_positions` (`index_copy_` / `scatter_`). Apply to `hidden_states`, `target_logprobs` (with `-100` label positions handled as today), `top_k`, and `logits`. `logits` are the one large case (seq × padded vocab); gathering them is still correct, but it may be worth keeping them shard-local behind an explicit opt-out since nothing in caladan requests full logits under CP.
3. Use an autograd-aware gather (the same pattern as `gather_from_sequence_parallel_region`, or `torch.distributed.nn.functional.all_gather`) so `hidden_states.requires_grad` losses computed on the gathered tensor route gradients back to the owning rank's rows. Because every CP rank then computes the same loss on the same full tensor, the reduce-scatter in the backward sums `cp_size` identical contributions; the gather's backward should scale by `1 / cp_size` (or the loss reduction should), matching whatever convention the existing TP sequence-parallel hidden gather uses for the probe head. Worth confirming that convention at the same time, since TP + `hidden_states` losses have the same replication.
4. Once outputs are replicated across the CP group, `dp_reduce` must not sum over CP ranks for quantities derived from full-sequence outputs. Today it reduces over `get_data_parallel_group(with_context_parallel=True)`, which is what makes shard-local sum losses correct; after this change the natural contract is `ForwardOutput` = full sequence and `dp_reduce` = pure DP (`with_context_parallel=False`). Any internal shard-sum path should reduce over the CP group itself before returning.

Net effect: CP becomes invisible to callers, as TP already is. This is the version we want; exposing positions and asking callers to map them is not acceptable as the fix.

## Environment

ART `bcfe13519` (caladan pins `a3a248a8e`), Megatron backend, dense `qwen3_5` handler (Qwen/Qwen3.8-27B, Qwen/Qwen3.6-27B class), single node H200:2 / H200:4 via caladan `Trainer`. Interim workaround: `ART_MEGATRON_CONTEXT_PARALLEL_SIZE=1` with TP = N (`ART_MEGATRON_TENSOR_MODEL_PARALLEL_SIZE`) or DP; measured on the 27B probe at 16 views/step: single GPU 118 s, TP 2 68 s, TP 4 40.5 s, DP 2 71 s.

bradhilton:
Context-parallel forwards returned rank-local sequence shards, so caller expressions such as `output.hidden_states[1:][mask]` failed with a full-sequence mask. All `ForwardOutput` fields now contain the full flattened input sequence in source order on every TP/CP rank. No public positions field or caller-side mapping is needed.

Internally, each CP rank scatters its owned rows into source order and sums across the CP group. This handles uneven, reordered, and empty shards without padded all-gather buffers. Backward routes one copy of the replicated loss gradient to each owning row, matching the existing TP sequence-parallel convention. `dp_reduce` now combines only DP batches; registered custom heads average replicated CP gradients while model/LoRA shard gradients still sum. Memory admission accounts for local and gathered outputs coexisting.

The public docs explicitly require identical losses across TP/CP replicas. The native correctness harness now uses the public forward API. GPU CI includes the new distributed regression cases and the updated custom-parameter reduction oracle.

Validation:
- Distributed/head suite: 16 passed, 1 skipped. Covers CPU CP2/CP4/DP2×CP2 and two-H200 NCCL CP2; full outputs, masked readouts, nonlinear losses, decoder/head/probe gradients, ignored labels, empty shards, frozen decoders, no-grad, and DP metrics. The four-GPU case skips on this two-GPU machine.
- The new CP2 regression fails against the unmodified parent at the full-sequence shape assertion.
- Native pretrained Qwen3-0.6B (one layer), CP2, two LoRA slots: all 16 output combinations pass across no-sharing/full-sharing layouts; head-chunk and slot-backward parity pass.
- Ruff, formatting, lockfile checks, and changed library/test type checks pass. Required `uv run prek run --all-files` reports the same 23 existing type diagnostics as the unmodified parent (exact comparison).
- Broader regression suite: 429 passed. Native GPU integration: 4 passed, 1 skipped (DP/TP/CP custom-head reduction and the TP head/backward oracle); the combined TP2×CP2 case requires four GPUs. Total: 449 selected tests passed, 2 skipped.
- Native long-context probe canary passed on two H200s: Qwen3.8-27B architecture with four randomly initialized layers, CP2/TP1, two 20,436-token shared-prefix views, 5,120-wide hidden states. The original `hidden_states[1:][mask]` expression works; 13,624 masked tokens are counted once, custom-head gradients match the analytical full-sequence reference, backward completes, and the optimizer updates the registered probe. This is an architecture/path canary, not a full pretrained 27B run.

The review follow-up guards the optional Megatron dependency before spawning distributed test workers. The backend-only CI environment now passes the 12 head tests and skips its five runtime-dependent cases; the same file with Megatron and H200s still passes 16 cases with only the four-GPU case skipped. The unguarded version was reproduced failing with `ModuleNotFoundError: megatron` in the backend environment.

Fixes #911.


bradhilton:
Minsky ACK: taking an independent full-head review of ART #912 at exact head `855736f3ef2a7af03a13edad5e7ac4d517b92356` on `49e33922` (OPEN, draft), requested by Brad. Scope: full source-order `ForwardOutput` fields on every TP/CP rank via per-rank scatter of owned rows plus CP-group sum (uneven, reordered and empty shards), backward routing of one replicated-loss gradient copy per owning row, `dp_reduce` combining only DP batches with registered custom heads averaging replicated CP gradients while shard gradients still sum, the memory-admission accounting for coexisting local and gathered outputs, the public docstring/export changes in `trainer_rank/__init__.py`, and the harness/CI/test changes. Explicit small/medium, public API, `art.megatron` and behavior assessment. Serial CPU checks only (threads 1, CUDA hidden, memory accounted; cgroup 298 GiB/512), including the CPU Gloo CP cases and a negative control against the base runtime; the H200 cases are the owner's. Verdict as a GitHub review after rechecking head AND state.

bradhilton:
McCarthy ACK: independent complete-diff source/CPU review at `855736f3ef2a7af03a13edad5e7ac4d517b92356` (merge-base `de95ed214c12c1892531c6478c7af847c65ebe69`, current main `49e33922`). I will check full-sequence reconstruction, autograd ownership, custom-parameter/DP reduction, empty shards and memory accounting. The public output/reduction semantics are intentional behavior changes and will be assessed explicitly against #911, separately from source correctness. Serial, bounded CPU checks only, native threads 1; no new delegates, provider/GPU workload, CI observer, shared environment edit, merge or deployment. Head will be rechecked before the verdict.

bradhilton (commented):
Review text removed on 2026-09-17; the consolidated review record is in the summary comment on this PR.

bradhilton:
McCarthy successor ACK: pre-post verification detected `1e068eed529cfb2cd18dcf8c0680c8c4b89d4297` and prevented posting the old-head verdict. I have checked the complete successor delta: two `megatron.core` pre-spawn test guards, one comment, and a clearer identical-replica-loss requirement. Runtime executable ASTs are unchanged from `855736f3ef2a7af03a13edad5e7ac4d517b92356`; the CPU results are retained at that original source, with an explicit equivalence receipt for this successor. Fresh exact-head verdict follows; no additional native rerun or resource action is needed for these non-executable runtime changes.

bradhilton (commented):
Review text removed on 2026-09-17; the consolidated review record is in the summary comment on this PR.

bradhilton:
@bradhilton McCarthy's fresh final-head source-correctness verdict is **CLEAR** at `1e068eed529cfb2cd18dcf8c0680c8c4b89d4297`: https://github.com/OpenPipe/ART/pull/912#pullrequestreview-5228075496. The pre-post head guard caught the update; the successor's doc/comment and two test dependency guards are independently reviewed, with executable-AST equivalence to the tested runtime recorded. CP2, empty-owner CP2 and DP2 CPU/Gloo controls pass across seven modes each; 13 adjacent tests pass. Earlier import failure/memory-guard stop remain preserved. Numerical/API-contract impact and CI/GPU limits are explicit. All review process groups are retired; no merge/deployment/adoption performed.

bradhilton (commented):
Review text removed on 2026-09-17; the consolidated review record is in the summary comment on this PR.

bradhilton:

**Change.** Context-parallel forwards returned rank-local shards, breaking full-sequence masks such as `hidden_states[1:][mask]`. All `ForwardOutput` fields now hold the full source-order sequence on every TP/CP rank via per-rank scatter plus CP-group sum; `dp_reduce` combines only DP batches and custom-head gradients average across CP. Fixes #911.

**Review.** Minsky and McCarthy independently cleared source correctness at `1e068ee` (an earlier unnamed review at `855736f`: LGTM pending GPU CI), both noting the CP semantics change is intentional, not behavior-neutral. Verified: CPU Gloo CP2/CP4/DP2xCP2 cases (64 passed, 2 skipped with Megatron; 12 passed, 5 skipped without), empty-shard and DP2 controls, negative control failing on the base runtime; two-H200 CI was pending at review time.

**Findings addressed during review.**
- CP tests failed with `ModuleNotFoundError: megatron` in the backend-only CI stage; `pytest.importorskip("megatron.core")` guards added.
- Docstring now requires identical losses on every TP/CP replica; comment distinguishes `dp_reduce` from internal CP-inclusive reductions.

**Deferred / follow-ups.**
- Four-GPU TP2xCP2 case unexercised; caladan 058 CP2/CP4 probe lanes to rerun.

Issue #913, opened 2026-09-16
Memory planner ignores recompute_granularity: selective recompute admitted a 32k-token group at 5.9 GB predicted, then OOM'd at >107 GB (TP 2/4, Qwen3.8-27B)


bradhilton:
## Summary

`TrainerRank.forward_micro_batches` admits a micro-batch after a memory-planner check, but the planner's activation estimate does not depend on `recompute_granularity`. With `ART_MEGATRON_RECOMPUTE_GRANULARITY=selective` (method/num_layers unset) on a 64-layer dense model (`Qwen/Qwen3.8-27B`, hidden 5120) the planner predicted **5.9 GB** for a ~32.7k-token packed group and admitted it; the forward then hit CUDA OOM against a **107–119 GB** budget on both TP 2 and TP 4:

```
art.trainer_rank.TrainerRankMemoryError: forward_micro_batches: CUDA OOM occurred despite the planner estimate.
packed_tokens=32710 logical_tokens=38443 predicted_peak_gb=5.894 usable_limit_gb=107.192.
Use smaller top-level items, reduce output requests, or call dp_rank_forward with already-DP-local smaller inputs.
```

(TP 4 rank: `predicted_peak_gb=5.894 usable_limit_gb=119.289`, same failure.) The OOM surfaces inside `_decoder_hidden` → `transformer_block.forward` → `_forward_mlp` → `art.megatron.lora.forward` on the very first forward of the run, so no learned memory profile exists yet and the static estimate is the only guard.

## Why the estimate is off

`_estimate_required_memory_bytes_from_values` (`src/art/trainer_rank/_impl.py`) uses

```
activation_factor = max(4, min(16, num_layers // 4 + 4))      # 16 for 64 layers
static_compute = packed_tokens * hidden_size * dtype_size * activation_factor
required = (output_bytes + static_compute) * 1.10
```

[...]

PR #914, opened 2026-09-16
Reject unsupported selective recompute in TrainerRank planning

bradhilton:
A selective-recompute runtime can pass TrainerRank's activation-memory admission check and then OOM on its first forward: the current static estimate is calibrated for full recompute. Reject `recompute_granularity="selective"` during TrainerRank construction with `TrainerRankRuntimeSupportError` and an actionable hint to restore full/uniform/one-layer recompute.

This is the first step for #913. Draft follow-up #915 models retained activations and removes the temporary selective-recompute refusal; it is stacked on this PR and awaits native GPU calibration. The provider and other Megatron entry points keep their existing behavior.

Validation:
- All 276 tests across topology, TrainerRank validation, and split admission passed (269 initially; seven multiprocessing cases passed after correcting the stdin-based test launcher).
- The selective-recompute rejection tests pass at TP 1/2/4 and all three fail against the original constructor. Full/disabled recompute acceptance is covered as well.
- Changed-file Ruff lint, formatting, type checks, and `uv lock --check` pass.
- Repository-wide type checking reports 23 existing diagnostics in untouched files.


bradhilton:
Minsky (independent reviewer): ACK — starting a full-head review of #914 at `8aa6d4f5ede453d1d90c7e511b2922b783c7a100`. Source and retained evidence first; bounded serial CPU checks only for concrete gaps; head and OPEN state re-verified before the verdict is posted.

bradhilton:
McCarthy ACK: independent full-diff source/CPU review at `8aa6d4f5ede453d1d90c7e511b2922b783c7a100` on `49e33922238126404248c4141adc32cdee9a8447`. I will verify the selective-mode refusal, its position before constructor side effects, preservation of PP errors and other modes, and the advertised recompute settings. I will keep this temporary refusal separate from #915's unqualified estimator and from general memory-safety claims. Serial native-thread-1 CPU checks only, one native test process at a time (about 1 GiB expected, 2 GiB hard process-tree guard; host about 262/400 GiB). No provider/GPU workload, dependency sync, new delegate/observer, shared edit, merge or deployment. Live head/state will be rechecked before posting.

bradhilton:
**Review (Shannon, posted as a comment because the gh token is Brad's account): LGTM as the stopgap for #913, with one gap worth naming.**

What I checked:
- Ran `tests/unit/test_trainer_rank_topology.py` against the PR head (`8aa6d4f5e`) in a scratch worktree: 14 passed, including the new `test_trainer_rank_recompute_support` cells at TP 1/2/4.
- The hint in the error matches the provider defaults exactly (`_apply_gpu_provider_defaults` sets `full` / `uniform` / `1`), so following it restores the calibrated configuration rather than some third state.
- The check reads `runtime.provider`, which `TrainingRuntime` carries after `_apply_runtime_env_overrides` + `_normalize_recompute_settings`, so it sees the effective value for the `ART_MEGATRON_RECOMPUTE_GRANULARITY=selective` path that produced the #913 OOM. It would have refused both failing runs at construction instead of on the first forward.

The gap: the guard rejects `"selective"` only, but the estimate is equally uncalibrated for `recompute_granularity=None` (no recompute at all), which is reachable two ways today: the `"none"` sentinel through the same env override, and `_enforce_ep_overlap_recompute_contract`, which deliberately sets `None` whenever `overlap_moe_expert_parallel_comm` is on. Both retain strictly more per-layer activation than `selective`, so a caller who reads this error and "fixes" it by setting `none` walks straight back into the same OOM. I would not widen the rejection to `None` here (EP overlap intentionally relies on it and presumably fits within its fixed budgets), but I'd add a sentence to the error that `none` is not the workaround, and make sure #915 prices `!= "full"` rather than `== "selective"`, which its diff does.

Minor, non-blocking: #915 will read the granularity from `metadata_model.config` with the provider as fallback; if that is the authoritative source it may be worth using the same lookup here so the two guards can't disagree when a model is built from a modified config.

bradhilton (commented):
Review text removed on 2026-09-17; the consolidated review record is in the summary comment on this PR.

bradhilton:
@bradhilton McCarthy source-correctness **CLEAR** at `8aa6d4f5ede453d1d90c7e511b2922b783c7a100`: https://github.com/OpenPipe/ART/pull/914#pullrequestreview-5228513339. The selective guard, PP-error precedence and nonselective behavior are verified (13 native constructor/topology cases; 72 additional source-boundary controls; base negative control 3 failed/6 passed). One nonblocking hint caveat concerns EP overlap normalizing full recompute to disabled. The unrelated local padding-test dependency failure and CI/qualification limits are retained explicitly. All review processes are retired; no merge/deployment/adoption performed.

bradhilton (commented):
Review text removed on 2026-09-17; the consolidated review record is in the summary comment on this PR.

bradhilton:

**Change.** Rejects `recompute_granularity="selective"` at TrainerRank construction with `TrainerRankRuntimeSupportError` and a hint to restore full/uniform/one-layer recompute: the static activation-memory admission estimate is calibrated for full recompute, so selective runtimes passed admission and OOMed on the first forward (#913). Other Megatron entry points are unchanged.

**Review.** Minsky and McCarthy posted source-correctness CLEAR at `8aa6d4f`; Shannon posted LGTM. Verified: topology tests pass (14/14, new selective-refusal cells at TP 1/2/4), the base fails exactly the three selective cases as a negative control, 72 further constructor controls pass, Ruff/format/ty clean. Hosted quality checks and the 2x H200 job were pending at review time.

**Deferred / follow-ups.**
- #915 (stacked draft) prices retained activations and removes this temporary refusal; not covered by these reviews.
- #913 remains only partially resolved: `None`/disabled recompute is still accepted with the same uncalibrated estimate.
- Non-blocking hint gaps: EP overlap normalizes full recompute to disabled; `none` is not a workaround; TrainerRank docstring omits the new refusal.

PR #915, opened 2026-09-16
Model retained activations and allow selective recompute in TrainerRank

bradhilton:
Selective and no-recompute execution retain decoder activations across layers, so the previous full-recompute heuristic admitted #913's long group and then OOMed. This follow-up to #914 replaces the selective refusal with a retained-activation floor that accounts for TP/SP, the actual attention/GDN layer mix, native MLP fusion, replicated QKV, prefix-segment states, and effective MLP/MoE checkpoints. Profiles can only raise the floor.

Native H200 calibration now puts the 27B estimate about 10–13% above measured peaks for the paired 2k-and-larger workloads. The previous TP4 estimate was about 60% high. Values are incremental allocated GiB; observed peaks are maxima across ranks and repetitions.

| Qwen3.8-27B workload | Previous estimate | New estimate | Observed forward |
| --- | ---: | ---: | ---: |
| TP4, 2k pair | 32.974 | 22.854 | 20.550 |
| TP4, 4k pair | 56.075 | 38.789 | 35.077 |
| TP4, 8k pair | 112.151 | 77.273 | 69.926 |
| TP4, 12k pair | — | 115.757 | 104.780 |
| TP8, original 19,221 + 19,222 pair | 153.866 (refused) | 126.188 | 114.665 |
| TP4, 8k pair with MLP recompute | 112.151 | 42.294 | 37.881 |

The previously unmeasured long TP8 case now completes forward and backward. TP4 still refuses the original long pair. Additional runs cover 4B/9B models, unfused eager/compiled execution, short batches with eight sequences, and MoE routing concentrated on one expert rank. MoE dispatch retains an imbalance allowance; the report documents its remaining conservatism and a native TE crash when expert peers are completely empty.

[Calibration report](https://github.com/OpenPipe/ART/blob/b86ad4bdd143e9d0a24e6204957c8ce970c20294/dev/trainer_rank_recompute_memory.md) · [CSV with source/driver hashes, peaks, and negative controls](https://github.com/OpenPipe/ART/blob/b86ad4bdd143e9d0a24e6204957c8ce970c20294/dev/trainer_rank_recompute_memory.csv)

Validation: 45 final GPU cells, 360 measured rank-samples and 4 refused rank-samples. Every measured forward is covered, every backward completes, and all measured losses/gradients are finite. Local TrainerRank tests: 631 passed, 2 skipped. Ruff and type checking pass. Full-recompute and no-grad behavior retain their previous estimates; the report explicitly limits the calibration's scope.

CI is green on `b86ad4bdd`: [quality checks](https://github.com/OpenPipe/ART/actions/runs/35167540749) (783 Megatron tests and 1,553 unit tests passed) and [native GPU validation](https://github.com/OpenPipe/ART/actions/runs/35167540822). Calibration clusters and the CI cluster were terminated successfully.

### Behavior with recompute disabled

The retained-activation floor also applies when `recompute_granularity=None`, including the `none` override and EP-overlap MoE trainers. Requests that previously fit the smaller heuristic may now be split or, for indivisible groups, refused. The MoE allowance protects uneven expert routing and can be roughly 3× the measured balanced-routing peak. Full recompute (the production default, including Caladan) and no-grad estimates are unchanged.

Context parallelism is not discounted in this version, so selective recompute on the default multi-GPU dense CP topology may be overestimated. CP scaling and GPU calibration are being addressed in a follow-up.


bradhilton:
**Review (Shannon; posted as a comment because the gh token is Brad's account): recommend merging.** The change is confined to the path that was previously refused, the calibration is real GPU evidence rather than CPU reasoning, and CI including the native GPU validation is green on `b86ad4bdd`.

What I verified in the diff and the report:
- The new floor is applied only when `signature.grad_enabled and recompute_granularity != "full"`, so the production default (`full`, set by `_apply_gpu_provider_defaults`) and every no-grad forward keep their existing estimate byte-for-byte; `test_full_recompute_keeps_existing_estimate` and `test_no_grad_does_not_pay_for_retained_layers` pin that.
- It prices `None` as well as `selective`, which closes the gap I raised on #914 (the `none` sentinel and the EP-overlap contract both land on `None`).
- Geometry lookup goes config → provider → default via `memory_field`, so the granularity, `sequence_parallel`, `attention_output_gate`, `bias_activation_fusion` and `recompute_modules` all come from the same source the model was built from. All `ModelGeometry` fields used (`gdn_conv_kernel`, `moe_experts`, `moe_shared_expert_ffn`, `num_query_groups`, …) already exist on `main`.
- Hybrid layer counts are priced separately (`_gdn_layers` vs attention), TP/SP sharding is applied to the per-layer terms with the gathered LoRA input kept unsharded, the replicated-QKV term for `num_query_groups < tp` explains the TP8 case, and the MLP checkpoint discount keeps one live workspace. Profiles can only raise the floor (`test_profile_cannot_erase_recompute_floor`).
- Calibration: 27B estimates sit 10–13% above measured peaks across 2k–19k-token pairs at TP4/TP8, minimum observed headroom 5.4%, every backward finite. The #913 32.7k-token pair is now refused at TP4 with a 181 GiB estimate instead of being admitted at 5.9 GB, and admitted at TP8 where it actually fits (114.7 GiB measured). That is exactly the behavior #913 asked for.
- The branch predates #920, so `git diff main` shows the fla_cp fix as reverted; the actual PR diff does not touch it and GitHub reports a clean merge, so no rebase is needed.

Two things worth recording as follow-ups (not blockers):
1. **CP is not modeled.** The floor reads `tp` from `_topology_key()` but never `cp`, and the token-proportional terms are not divided by the context-parallel size, while the report explicitly scopes CP out of the calibration. Under ART's default dense topology on 2+ GPUs (TP 1, CP = visible GPUs) the retained-activation floor is therefore ~CP× too high, so selective recompute will be refused or split more than necessary there. Conservative, so safe to ship, but it means the practical way to use selective recompute on multi-GPU today is TP (as calibrated), not the default CP. A follow-up that divides the per-token terms by `cp` (GDN recurrent-state terms scale with segments, not tokens, and should stay as is) plus one CP calibration cell would close it.
2. **Behavior change for the `None` path.** Any runtime that ran with recompute disabled (EP-overlap MoE trainers, or the `none` override) now gets a floor that the report describes as deliberately loose for balanced routing (~3× the balanced peak at 1.5k–6k tokens on 35B-A3B). Workloads that fit before may now be split or, for indivisible groups, refused. Caladan is unaffected (all our trainers are on `full`), but it would be good for the PR description or changelog to state this so an EP-overlap user reading a new refusal knows why.

Minor: the `dev/` CSV/report/driver/SkyPilot task are a useful record; the reproduction commands and the explicit limits paragraph ("does not validate … CP, pretrained routing, larger LoRA ranks") are what make this reviewable. Thanks for the thoroughness.

bradhilton:
Addressed both follow-ups. #915 is merged, and its description now explicitly covers the behavior change for `None`, the `none` override, and EP-overlap trainers.

CP support is in #922 with native H200 calibration: 13 cells / 52 rank-samples, every forward covered and every backward finite. The 27B 4k pair drops from 117.4 GiB estimated to 69.5 GiB against 62.9 GiB observed and now fits.

One refinement to the suggested division: ART's actual CP ownership is uneven. That pair packs 6,964 tokens but assigns 4,096 to its busiest rank; dividing by two would miss the measured peak. The follow-up uses the native planner's maximum attention/GDN rank load, preserves the segment-state charge, and keeps MoE's dispatch allowance until it has its own CP calibration. Local tests pass (637, with 2 skipped); final CI is running on #922.

bradhilton:

**Change.** Replaces TrainerRank's selective-recompute refusal with a retained-activation floor accounting for TP/SP, attention/GDN layer mix, MLP fusion, replicated QKV, prefix-segment states and MLP/MoE checkpoints, after the full-recompute heuristic admitted #913's long group and OOMed (follow-up to #914). H200 calibration puts 27B estimates 10–13% above measured peaks; the long TP8 pair now fits.

**Review.** Shannon reviewed and recommended merging at `b86ad4b`, confirming the floor applies only to grad-enabled non-full recompute (full-recompute and no-grad estimates unchanged), prices `None` as well as `selective`, and that profiles can only raise it. Verified: 631 local TrainerRank tests passed (2 skipped), ruff and type checks, green CI (783 Megatron and 1,553 unit tests plus native GPU validation), 45 GPU calibration cells.

**Deferred / follow-ups.**
- Context parallelism not modeled, so selective recompute on the default multi-GPU CP topology is overestimated; fixed in #922 with CP calibration.
- Behavior change for recompute-disabled (`None`) paths, including EP-overlap MoE trainers, documented in the PR description after review.

